PR #28, opened 2015-07-31
Fixed drm issue.

yohanesu:
Seems to be working fine however in my UI animations it seems like it doesn't clear the background properly before rendering at at the new location (running in window under X). Didn't have this problem with software rendering. Will try run natively and see if there is any difference.

### DIFF
--- a/src/platform/context/egl.c
+++ b/src/platform/context/egl.c
@@ -54,6 +54,7 @@ static struct {
       EGLBoolean (*eglMakeCurrent)(EGLDisplay, EGLSurface, EGLSurface, EGLContext);
       EGLBoolean (*eglSwapBuffers)(EGLDisplay, EGLSurface);
       EGLBoolean (*eglSwapInterval)(EGLDisplay, EGLint);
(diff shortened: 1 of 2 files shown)


yohanesu:
Hmm there's something weird but I don't know if it the compositor or cairo-gl. I'm animating some balls across the screen. For one second it works as expected (background gets cleared at old position and balls rendered at new position). The the other second the screen flickers and the background isn't cleared properly. Edges of the balls remains as a trail. Did not have this problem in shm mode. The switching is extremely periodic, 1 sec ok, 1 sec flicker and repeating...  If I clear the whole background each frame it works fine.

Cloudef:
Does this PR make wl_drm show up in registry as well?

yohanesu:
Move to egl.c: Yeah I figured it belong there but not sure how to share it to gles2.c. I'll fix it according to your instructions today or tomorrow. 
wl_drm: Yes, it shows up in registry and hardware rendering works in window under X11 and natively for my clients.

yohanesu:
I was looking how weston solved it and did similar solution.

Cloudef:
I've pushed similar fix to wl_drm branch, we can close this if it's fine.

PR #33, opened 2015-08-10
Improve documentation of wlc interface

### LINKED ISSUE #29: Insufficent documentation for keyboard.key
What is the bool return value supposed to be?

https://github.com/Cloudef/wlc/blob/master/include/wlc/wlc.h#L133

SirCmpwn:
Fixes #29.

As an aside, it'd be nice if we could simulate these events.

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -129,23 +129,23 @@ struct wlc_interface {
    } view;
 
    struct {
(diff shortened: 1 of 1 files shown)


Cloudef:
Thank you.

I think it might be cleaner to write component that simulates the events and sends them through wlc_interface compatible structs, so you can just swap the implementation.

PR #43, opened 2015-08-18
Don't hardcode installation of libraries to "lib"

lkundrak:
It's a FHS violation on PPC64, s390x, sparc64 and AMD64.


Cloudef:
http://www.cmake.org/cmake/help/v3.3/module/GNUInstallDirs.html This seems to be what we want.

PR #44, opened 2015-08-18
Fix order in which privileges are dropped

lkundrak:
See: POS36-C. Observe correct revocation order while relinquishing privileges


Cloudef:
Thank you!

PR #45, opened 2015-08-18
xwm: attempt to load libxcb-image.so.0 first

lkundrak:
It makes more sense than the .so file: it's a known SONAME that ought to
contain the ABI we're looking for and .so file is not going to be present for
installations without development packages.


lkundrak:
Uh, oh, this is done in more places too. Not good for Linux distribution packaging at all.

Let's try another approach, pull request #46

Cloudef:
I'll close this then.

PR #46, opened 2015-08-18
Make it possible to disable weak linking

lkundrak:
It's rather fragile -- the .so files provide no ABI stability whatsoever
(unlike relying on SONAMEs). This also makes it cumbersome to intergate RPM
packaging as it effectively defeats the dependency generators as no .dynsym
entries are generated.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -130,6 +130,26 @@ endif ()
 
 set_target_properties(wlc PROPERTIES VERSION ${WLC_VERSION} SOVERSION ${SOVERSION})
 
+if (NO_WEAK_LINK)
+   list(APPEND definitions -DNO_WEAK_LINK)
+   list(APPEND libraries
+      ${GLESv2_LIBRARIES}
+      ${GBM_LIBRARIES}
+      ${DRM_LIBRARIES}
+      ${X11_LIBRARIES}
+      ${X11_XCB_LIBRARIES}
+      ${XCB_LIBRARIES_xkb}
+      ${XCB_LIBRARIES_composite}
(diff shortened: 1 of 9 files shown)


lkundrak:
Also, this generates a couple of mismatch warnings for the NO_WEAK_LINK case.

It has aleady been wrong, but the dlopen() dance effectively defeated the compiler's type checking. :(

Cloudef:
Better solution would be to compile plugins and link against the plugins. I wanted to avoid this to keep the library itself simple though.

robotanarchy:
I ran into the same issue while porting `wlc` to Void Linux.

[My workaround](https://github.com/robotanarchy/void-packages/commit/f3ad55ca2a9e0002f82bb14c627f1d9289964396): Before compiling, replace all `"lib*.so"` strings with their correct soname by resolving the links in `/usr/lib` in the build environment.

```bash
echo "Replacing dynamically loaded libraries with their sonames..."
while IFS= read -r i; do
	file="$(echo $i | cut -d ':' -f 1)"
	lib="$(echo $i | cut -d '"' -f 2)"
	[ -e "/usr/lib/$lib" ] || continue
	lib_soname="$(basename $(realpath /usr/lib/$lib))"
	echo "$file: $lib -> $lib_soname"
	sed --in-place "s/\"$lib\"/\"$lib_soname\"/g" "$file"
done < <(grep -r -E '"lib.+\.so"' --only-matching src)
```

As an alternative to the plugins architecture, how about putting the correct sonames with CMake in a generated header file and include that headerfile where necessary?

Cloudef:
285ec2d5e24e1c0a6768b16592fe4bf3b416c109

PR #49, opened 2015-08-18
Make CMake fail if wayland is not recent enough

hsoft:
Without this explicit failure, CMake goes through fine, but we get a
cryptic compilation error during `make`.

Background: I'm on Gentoo and the current stable wayland version is 1.6.1. It took me a while to figure out why wlc wouldn't compile.


Cloudef:
wlc is in development, I don't guarantee API or ABI stability at this point. Nor supporting running at non development version of wayland or xdg-shell at this point.

Issue #47, opened 2015-08-18
Separate optional components to plugins


Cloudef:
Link against these plugins and dlopen the plugins instead. Just like orbment / bemenu does.

Cloudef:
#46

Cloudef:
Started working on feature/modular branch.

PR #68, opened 2015-09-04
Properly detect and handle includedirs

scarabeusiv:
This avoids buildfails in some cases like when building for openSUSE.

Overall the find_package were moved to main subfolder where all the
detected variables are later on propagated for each subdir cmakelists.

There we just add the required -I paths as needed.


Cloudef:
Looks good. I'll test later today and merge then.
FYI, see also this relevant issue for removing weak linkage https://github.com/Cloudef/wlc/issues/47

PR #74, opened 2015-10-22
Add wlc_pointer_set_origin

### LINKED ISSUE #71: manipulating mouse-position
Hi,

As far as I can see, there is no feature to manipulate the mouse position. This would be nice to have for a virtual desktop size (for example, as it has been implemented in waimea).
I just browsed the sourcecode and in general it is possible to implement this with WLC. But it will certainly change the API of WLC, depending on the implementation massively.

Is such a feature wanted, or are there objections against such a feature?

SirCmpwn:
Closes #71


Cloudef:
#71 proposed interface with output, but you get same effect by focusing output. For now lets merge this and see how and where it's used. Thanks.

PR #75, opened 2015-10-22
Move responsibility for updating pointer to client

SirCmpwn:


Cloudef:
Thanks. /me updates orbment

PR #76, opened 2015-10-23
Rename git submodule to be easier to set

linkmauve:
This makes the AUR package use the correct path, and probably helps other users as well.


Cloudef:
We tried to do this before, but it broke something. I don't remember what anymore. But I should check my irc logs. There is probably good reason git defaults to full path here though.

Issue #77, opened 2015-10-27
Crash in wlc on sway startup


SirCmpwn:
Error: Signal 6. Printing backtrace
    Backtrace: /home/sircmpwn/sources/sway/bin/sway(error_handler+0x9f) [0x410af2]
    Backtrace: /usr/lib/libc.so.6(+0x33680) [0x7ff70acbe680]
    Backtrace: /usr/lib/libc.so.6(gsignal+0x38) [0x7ff70acbe5f8]
    Backtrace: /usr/lib/libc.so.6(abort+0x16a) [0x7ff70acbfa7a]
    Backtrace: /usr/lib/libwlc.so.0(+0x15d6c) [0x7ff70b6ffd6c]
    Backtrace: /usr/lib/libwlc.so.0(+0x163b2) [0x7ff70b7003b2]
    Backtrace: /usr/lib/libwlc.so.0(+0x17a19) [0x7ff70b701a19]
    Backtrace: /usr/lib/libwlc.so.0(+0x18769) [0x7ff70b702769]
    Backtrace: /usr/lib/libwlc.so.0(+0x9795) [0x7ff70b6f3795]
    Backtrace: /usr/lib/libwlc.so.0(+0x8698) [0x7ff70b6f2698]
    Backtrace: /usr/lib/libwlc.so.0(+0x11e0c) [0x7ff70b6fbe0c]
    Backtrace: /usr/lib/libwlc.so.0(+0x7d8f) [0x7ff70b6f1d8f]
    Backtrace: /usr/lib/libwlc.so.0(+0x1dcfc) [0x7ff70b707cfc]
    Backtrace: /usr/lib/libwlc.so.0(wlc_run+0x55) [0x7ff70b708075]
    Backtrace: /home/sircmpwn/sources/sway/bin/sway(main+0x274) [0x41055a]
    Backtrace: /usr/lib/libc.so.6(__libc_start_main+0xf0) [0x7ff70acab610]
    Backtrace: /home/sircmpwn/sources/sway/bin/sway(_start+0x29) [0x404379]

Cloudef:
Can I have more detailed trace. Either missing debugging symbols or the backtrace function in glibc is not good enough as gdb.

SirCmpwn:
Yeah, I'll get you a better stack when I could get a chance.

Cloudef:
Thanks

PR #78, opened 2015-10-28
gles2: Fix error with ortho initialisation.

Earnestly:
The use of non-constant initialisation in the global scope is not supported by GLSL/GLES, this patch moves the initialisation of ortho into main instead.

Fixes an issue with recent mesa which now properly enforces this property:

```
(ERROR) wlc: Compiling vertex: 0:4(14): error: initializer of global variable `ortho' must be a constant expression
```


SirCmpwn:
Fixes #77

Cloudef:
Thanks!

PR #81, opened 2015-11-04
Fixed window naming in X11 backend

yohanesu:
Since we use xcb calls I changed to XCB_ATOM_*. Window name now showing correctly in the window frame and with wmctrl -l.


yohanesu:
Forgot window class.. Making a new PR.

Cloudef:
Thanks

PR #84, opened 2015-11-11
example: Update/fix, set pointer origin.

sce:
Poor mouse pointer, stuck in the top left corner...


Cloudef:
Indeed, I have forgotten this. Thanks

PR #92, opened 2015-12-02
Link to DBus and systemd

gpyh:
DBus and systemd were not linked are compile time. Now they are.
In most cases, wlc works without this change, but it's because if DBus and systemd were found, it usually means that the linker will be able to find them at runtime.
This is not the case on the NixOs Linux distribution, which clears LD_LIBRARY_PATH after building and solely uses the binairies and libraries' RPATH to find the dynamic libraries. If a library is not linked at compile time (although still *dynamically* linked and not statically linked), the RPATH is not properly written.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -111,6 +111,34 @@ target_link_libraries(wlc
    ${CMAKE_DL_LIBS}
    )
 
+if (DBUS_FOUND)
+  target_link_libraries(wlc
+    PRIVATE
+    ${DBUS_LIBRARIES}
+    )
+endif()
+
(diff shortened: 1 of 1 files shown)


Cloudef:
See this https://github.com/Cloudef/wlc/pull/46
And this https://github.com/Cloudef/wlc/issues/47

gpyh:
Is the feature/modular branch usable?
If so, I'll pull this one ; otherwise I will package wlc with a patch until this is implemented.

I would be happy to help making this happen.

Cloudef:
feature/modular is bit of behind, and not in usable state. I suggest packaging with the patch until implemented.

PR #96, opened 2015-12-21
Close xwayland before the compositor

crondog:
In sway if you have an xwayland window open eg a terminal and then exit
sway without closing this terminal, wlc crashes and leaves sock files
behind


Cloudef:
Makes sense.

PR #101, opened 2016-01-03
Update keyboard leds when pressed

crondog:
Please let me know if you have a better way of doing this. This was the easiest/minimal changes that I could think of

### DIFF
--- a/src/compositor/seat/keyboard.c
+++ b/src/compositor/seat/keyboard.c
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <wayland-server.h>
+#include <libinput.h>
 #include "internal.h"
 #include "macros.h"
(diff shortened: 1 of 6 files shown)


Cloudef:
This looks good. wlc_keyboard_update_leds could take bitmask instead of the wlc_keyboard struct though. You can also drop wlc_ prefix and remove the function prototype from keyboard.h and declare the function simply as static in keyboard.c since that's the only place it's used from.

crondog:
Thanks for the suggestions. That is much better

Cloudef:
Thanks, I'll give it a go later today.

crondog:
Hmm so I received an email saying add an assert, but now I can't see it on gh. I think I added in the right place...

Cloudef:
Yup, that's correct. Thanks

PR #105, opened 2016-01-06
Fix touchscreen crashes

crondog:
NOTE: This does not get touch working. Only MOTION and DOWN events are
sent to a client. This is because wlc_touch_touch tries to find the
underlying view via view_under_touch which fails for UP, FRAME, CANCEL since
they do not send the pos, so we return before
wl_touch_send_up,frame,cancel can be called.

### DIFF
--- a/src/compositor/seat/touch.c
+++ b/src/compositor/seat/touch.c
@@ -75,7 +75,7 @@ wlc_touch_touch(struct wlc_touch *touch, uint32_t time, enum wlc_touch_type type
          case WLC_TOUCH_DOWN:
          {
             uint32_t serial = wl_display_next_serial(wlc_display());
-            wl_touch_send_down(wr, serial, time, surface, slot, wl_fixed_from_double(pos->x), wl_fixed_from_double(pos->y));
+            wl_touch_send_down(wr, serial, time, surface, slot, wl_fixed_from_int(pos->x), wl_fixed_from_int(pos->y));
          }
          break;
 
@@ -87,7 +87,7 @@ wlc_touch_touch(struct wlc_touch *touch, uint32_t time, enum wlc_touch_type type
          break;
 
          case WLC_TOUCH_MOTION:
-            wl_touch_send_motion(wr, time, slot, wl_fixed_from_double(pos->x), wl_fixed_from_double(pos->y));
+            wl_touch_send_motion(wr, time, slot, wl_fixed_from_int(pos->x), wl_fixed_from_int(pos->y));
             break;
 
          case WLC_TOUCH_FRAME:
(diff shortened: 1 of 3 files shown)


src/compositor/seat/touch.c
```diff
@@ -87,7 +87,7 @@ wlc_touch_touch(struct wlc_touch *touch, uint32_t time, enum wlc_touch_type type
          break;
 
          case WLC_TOUCH_MOTION:
-            wl_touch_send_motion(wr, time, slot, wl_fixed_from_double(pos->x), wl_fixed_from_double(pos->y));
+            wl_touch_send_motion(wr, time, slot, pos->x, pos->y);
             break;
```
Cloudef: These functions take wl_fixed as argument, hence wl_fixed_from_double is needed.

src/session/udev.c
```diff
@@ -200,10 +200,22 @@ input_event(int fd, uint32_t mask, void *data)
          break;
 
          case LIBINPUT_EVENT_TOUCH_UP:
+         {
+            struct libinput_event_touch *tev = libinput_event_get_touch_event(event);
+            struct wlc_input_event ev;
```
Cloudef: zero initialize this struct ev = {0}; instead and remove the = NULL's below

src/session/udev.c
```diff
@@ -212,11 +224,28 @@ input_event(int fd, uint32_t mask, void *data)
...
             ev.touch.slot = libinput_event_touch_get_seat_slot(tev);
             wl_signal_emit(&wlc_system_signals()->input, &ev);
          }
          break;
 
+         case LIBINPUT_EVENT_TOUCH_FRAME:
+         case LIBINPUT_EVENT_TOUCH_CANCEL:
+         {
+            struct libinput_event_touch *tev = libinput_event_get_touch_event(event);
+            struct wlc_input_event ev;
```
Cloudef: Same here.

src/session/udev.c
Cloudef: Probably good idea to do the zero intialize for every input event here. But better do that as another commit/PR to keep focus.

src/compositor/seat/touch.c
crondog: struct wlc_point {
   int32_t x, y;
};
These don't look like doubles to me

Cloudef: wl_fixed isn't int32_t either, though you should change wl_fixed_from_double to wl_fixed_from_int here, it's binary scaled unit (truncated using division and expanded using multiplication, don't ask the wl people were on acid when they did it and how they implemented it is even more crazy)

That implicit double conversion most likely is just copy paste from pointer.c that does use double for x and y, unlike here.

crondog: My eyes. My brain.

Cloudef:
Thanks, I can't test this with actual hardware but it looks fine.

PR #107, opened 2016-01-14
Fix repeat keys for WM commands.

mikkeloscar:
Makes sure that key combos/wm commands grapped by the compositor is
repeated by wlc.


Cloudef:
Thanks!

PR #110, opened 2016-01-18
gitignore

crondog:
Just to make things nicer and not accidentally add all the things

I think the same needs to be done to chck to stop it showing in wlc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,17 @@
+*.a
+*.pc
+*.so
+*.so.*
+*.swp
+.ycm_extra_conf.py*
+CMakeCache.txt
+CMakeFiles
+CPackConfig.cmake
+CPackSourceConfig.cmake
+CTestTestfile.cmake
+DartConfiguration.tcl
+Makefile
+cmake_install.cmake
+example/example
+tests/resources-test
+wayland-*-protocol.*


Cloudef:
It might be better to just add * as ignore and add files with -f option. The .gitignore will otherwise just becomes big mess.

crondog:
Well it's up to you. I feel like there might not be any more than this...Unless a lot of weird file types are going to be added

Cloudef:
Well tests/ folder will pile up with binaries, and ignoring tests folder will ignore both src and binary folder. It's sort of inconvenient. In the end you get same result by making ignores opt-out rather opt-in.

mikkeloscar:
It is possible to ignore binaries with something like:

```
tests/*
!tests/*.*
```

Cloudef:
It will also ignore the source files

mikkeloscar:
No, actually, the first line ignores everything, and then the second line "un-ignores" files with a dot/extension e.g. source files. This just has to be put before the cmake ignores in `.gitignore` otherwise they will get un-ignored as well.

Cloudef:
All right, but I still don't see the point of doing this trickery. When you can simly ignore everything and add files with -f, you can totally forget .gitignore by doing this.

mikkeloscar:
I think it's just convenience. I personally use `git status` a lot and it is very helpful when it only shows the relevant files and folders and hides everything else.

But I'm not trying to change your mind, just wanted to point out that it is possible to create a somewhat static `.gitignore` that ignores everything it should, without needing to modify it in the future (unless the project structure changes). :)

Cloudef:
git status will show anything you've added to the git index regardless of .gitignore.

Earnestly:
The sane solution to this is making git default to ignoring everything by default via `$GIT_DIR/info/exclude => *` ensuring that only the files added (whitelisting) via `git add -f` are tracked.

But no, we live in a world where `.gitignore` is considered sane, [yes, sure](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition/blob/master/.gitignore).

Cloudef:
Rather than accepting this PR, I would suggest people that want to use .gitignore, just put their target folder they build wlc in locally to .gitignore. That way you make sure you only have the source files marked as untracked. 

This doesn't make sense in wlc for three reasons 1. I don't dictate where you build wlc, example in readme says `target` but that's just example. 2) I personally just ignore everything. 3) It adds yet another unrelated meta dotfile into the repository which I want to keep low as possible.

Editor / developer specific files should also be in your own global git ignore and not in project specific gitignore.

PR #111, opened 2016-01-24
Fix qt menus

crondog:
Clicking a menu in texstudio in sway results in the menu being treated
as a view and it gets tiled. This fixes that


SirCmpwn:
:+1:

crondog:
Shit. This breaks FF. As it no menus open...

crondog:
Well I'll just close this then. Should have tested it more thoroughly

Cloudef:
Yeah WINDOW_TYPE_NORMAL should not ever be a menu..

PR #114, opened 2016-01-28
Add simple .gitignore

SirCmpwn:
When building according to the readme, build artifacts are all under
target/. This change omits that directory from version control.


Cloudef:
This is better locally as I personally have way more targets than one target/. I would expect every people have slightly different looking local copy of the repository. If they are serious about contributing it doesn't take long to `echo target/ > .gitignore`

SirCmpwn:
If they do that, they'll check their .gitignore in, too. This is a simple change, hurts no one, and helps contributors.

SirCmpwn:
Remember as well that not all contributors are long term ones. They may just clone the repo, build it according to the readme, make a small fix, and send a PR.

PR #116, opened 2016-01-31
wlc_interface: Add compositor.terminate

sardemff7:


src/compositor/compositor.c
```diff
@@ -237,6 +237,7 @@ terminate_event(struct wl_listener *listener, void *data)
    (void)data;
    struct wlc_compositor *compositor;
    except(compositor = wl_container_of(listener, compositor, listener.terminate));
+   WLC_INTERFACE_EMIT(compositor.terminate);
```
Cloudef: Can you move this into body of wlc_compositor_terminate instead (right after the wlc_log "Terminating compositor..." and state variable change)

PR #120, opened 2016-02-13
Implement subsurfaces

### LINKED ISSUE #22: Implement subsurfaces
We implement the interface, but it's actually stubbed.

ammen99:
Should resolve #22 

This is my work on subsurfaces inside WLC. I haven't tested them a lot(the only compositor tested is orbment), but gedit and nautlilus work properly now(and they use subsurfaces). If there is anything which doesn't work, please tell me, I'll do my best to fix it.

What this pull request does is rendering the subsurfaces of a surface and also modify pointer focus functions to support focusing of subsurfaces. 

Things that still need to be done(aka TODO list):

1. Implement proper handling of synchronized subsurfaces - DONE
2. Implement subsurface::place_above and subsurface::place_below - DONE, but untested

### DIFF
--- a/src/compositor/output.c
+++ b/src/compositor/output.c
@@ -234,6 +234,43 @@ finish_frame_tasks(struct wlc_output *output)
 }
 
 static void
+render_subsurface(struct wlc_output *output, struct wlc_surface *surface, struct wlc_point offset, struct wlc_coordinate_scale parent_scale)
+{
+   const struct wlc_geometry g = (struct wlc_geometry) {
+       .origin = {offset.x + parent_scale.w * (surface->commit.subsurface_position.x + surface->commit.offset.x),
+                  offset.y + parent_scale.h * (surface->commit.subsurface_position.y + surface->commit.offset.y)},
+       .size = surface->size
+   };
+   wlc_render_surface_paint(&output->render, &output->context, surface, &g);
+}
+
+static void
+subsurfaces_render(struct wlc_output *output, struct wlc_surface *surface, struct wlc_coordinate_scale parent_scale, struct chck_iter_pool *callbacks, struct wlc_point offset)
+{
+
+   if (!surface)
+       return;
+   /* do not render view's main surface twice */
+   if (surface->parent)
+       render_subsurface(output, surface, offset, parent_scale);
+
+   wlc_resource *sub;
+   chck_iter_pool_for_each(&surface->subsurface_list, sub) {
+       subsurfaces_render(output, convert_from_wlc_resource(*sub, "surface"), surface->coordinate_transform,
+             callbacks,
+             (struct wlc_point) {
+              offset.x + (surface->parent ? 0 : surface->commit.subsurface_position.x / parent_scale.w),
+              offset.y + (surface->parent ? 0 : surface->commit.subsurface_position.y / parent_scale.h)
+             });
+   }
+
+   wlc_resource *r;
+   chck_iter_pool_for_each(&surface->commit.frame_cbs, r)
+      chck_iter_pool_push_back(callbacks, r);
+   chck_iter_pool_flush(&surface->commit.frame_cbs);
+}
+
+static void
 render_view(struct wlc_output *output, struct wlc_view *view, struct chck_iter_pool *callbacks)
 {
    assert(output && callbacks);
@@ -249,13 +286,13 @@ render_view(struct wlc_output *output, struct wlc_view *view, struct chck_iter_p
    WLC_INTERFACE_EMIT(view.render.pre, convert_to_wlc_handle(view));
    wlc_render_flush_fakefb(&output->render, &output->context);
    wlc_render_view_paint(&output->render, &output->context, view);
+
+   struct wlc_geometry b;
+   wlc_view_get_bounds(view, &b, NULL);
+   subsurfaces_render(output, surface, (struct wlc_coordinate_scale) {1, 1}, callbacks, b.origin);
+
    WLC_INTERFACE_EMIT(view.render.post, convert_to_wlc_handle(view));
    wlc_render_flush_fakefb(&output->render, &output->context);
-
-   wlc_resource *r;
-   chck_iter_pool_for_each(&surface->commit.frame_cbs, r)
-      chck_iter_pool_push_back(callbacks, r);
-   chck_iter_pool_flush(&surface->commit.frame_cbs);
 }
 
 static bool
--- a/src/compositor/seat/pointer.c
+++ b/src/compositor/seat/pointer.c
@@ -63,13 +63,46 @@ view_visible(struct wlc_view *view, uint32_t mask)
    return (view->mask & mask);
 }
 
-static struct wlc_view*
-view_under_pointer(struct wlc_pointer *pointer, struct wlc_output *output)
+static void
+find_surface_at_position_recursive(struct wlc_pointer *pointer, struct wlc_surface *parent, struct wlc_focused_surface *out)
 {
-   assert(pointer);
+   wlc_resource *sub;
+   chck_iter_pool_for_each(&parent->subsurface_list, sub) {
+      struct wlc_surface *subsurface;
+      if (!(subsurface = convert_from_wlc_resource(*sub, "surface")))
+         continue;
+
+      int32_t dx = subsurface->commit.subsurface_position.x * parent->coordinate_transform.w;
+      int32_t dy = subsurface->commit.subsurface_position.y * parent->coordinate_transform.h;
+
+      out->offset.x += dx;
+      out->offset.y += dy;
+      find_surface_at_position_recursive(pointer, subsurface, out);
+
+      if (out->id)
+         return;
+
+      if (out->offset.x <= pointer->pos.x && out->offset.y <= pointer->pos.y &&
+            subsurface->size.w + out->offset.x >= pointer->pos.x &&
+            subsurface->size.h + out->offset.y >= pointer->pos.y) {
+         out->id = *sub;
+         return;
+      }
+
+      out->offset.x -= dx;
+      out->offset.y -= dy;
+   }
+}
+
+static bool
+surface_under_pointer(struct wlc_pointer *pointer, struct wlc_output *output, struct wlc_focused_surface *out)
+{
+   assert(pointer && out);
 
    if (!output)
-      return NULL;
+      return false;
+
+   out->id = 0;
 
    wlc_handle *h;
    chck_iter_pool_for_each_reverse(&output->views, h) {
@@ -79,13 +112,24 @@ view_under_pointer(struct wlc_pointer *pointer, struct wlc_output *output)
 
       struct wlc_geometry b;
       wlc_view_get_bounds(view, &b, NULL);
-      if (pointer->pos.x >= b.origin.x && pointer->pos.x <= b.origin.x + (int32_t)b.size.w &&
-          pointer->pos.y >= b.origin.y && pointer->pos.y <= b.origin.y + (int32_t)b.size.h) {
-         return view;
+
+      struct wlc_surface *surface;
+      if ((surface = convert_from_wlc_resource(view->surface, "surface"))) {
+         out->offset = b.origin;
+         find_surface_at_position_recursive(pointer, surface, out);
+
+         if (out->id) {
+            return true;
+         } else if (b.origin.x <= pointer->pos.x && b.origin.y <= pointer->pos.y &&
+                    surface->size.w + b.origin.x >= pointer->pos.x &&
+                    surface->size.h + b.origin.y >= pointer->pos.y) {
+            out->id = view->surface;
+            return true;
+         }
       }
    }
 
-   return NULL;
+   return false;
 }
 
 static void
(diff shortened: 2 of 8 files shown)


Cloudef:
If you haven't seen it already https://github.com/Cloudef/wlc/blob/master/CONTRIBUTING.rst this file goes over the general code styling quite well as FYI. But this code overall looks good.

crondog:
I just tested this with weston-subsurfaces and the pointer disappears when going inside the subsurfaces (the 2 small squares)

ammen99:
Yes, you're right. I'll investigate that.

ammen99:
The last commit should resolve the disappearing cursor.

crondog:
The pointer is now visible. However the subsurfaces don't appear until I put the pointer inside the application.

Cloudef:
>The pointer is now visible. However the subsurfaces don't appear until I put the pointer inside the application.

The code probably needs wlc_output_schedule_repaint when surface is added. It might be better to do this in wlc_surface_add_child_surface(parent, child) where output from parent is copied to child and that output is also scheduled.

ammen99:
Possible reason for this bug is that weston-subsurfaces switches between synchronized and desynchronized mode, which is still to be implemented(but I don't know for sure).

src/compositor/output.c
```diff
@@ -234,6 +234,44 @@ finish_frame_tasks(struct wlc_output *output)
 }
 
 static void
+render_subsurface(struct wlc_output *output, struct wlc_surface *surface, struct wlc_point offset, struct wlc_coordinate_scale parent_scale) {
+   struct wlc_geometry g = (struct wlc_geometry) {
```
Cloudef: this looks like candicate for const

src/compositor/output.c
```diff
@@ -234,6 +234,44 @@ finish_frame_tasks(struct wlc_output *output)
...
+       .size = surface->size
+   };
+
+   wlc_render_surface_paint(&output->render, &output->context, surface, &g);
+}
+
+static void
+subsurfaces_render(struct wlc_output *output, struct wlc_surface *surface, struct wlc_coordinate_scale parent_scale, struct chck_iter_pool *callbacks, struct wlc_point offset) {
+
+   if(!surface)
```
Cloudef: Separate those poor if's from the parenthesis :P

src/compositor/output.c
```diff
@@ -234,6 +234,44 @@ finish_frame_tasks(struct wlc_output *output)
...
+       return;
+
+   /* do not render view's main surface twice */
+   if(surface->parent)
+       render_subsurface(output, surface, offset, parent_scale);
+
+   wlc_resource *sub;
+
```
Cloudef: Useless whitespace

src/compositor/seat/pointer.c
```diff
@@ -63,13 +63,43 @@ view_visible(struct wlc_view *view, uint32_t mask)
...
+      if(!subsurface) {
+         wlc_log(WLC_LOG_ERROR, "Invalid subsurface id %" PRIuWLC " parent %" PRIuWLC,
+                 convert_to_wlc_resource(subsurface), convert_to_wlc_resource(parent));
+      }
+
+      if(subsurface->commit.subsurface_position.x * parent->scale.w + out->offset.x <= pointer->pos.x &&
+         subsurface->commit.subsurface_position.y * parent->scale.h + out->offset.y <= pointer->pos.y &&
+         subsurface->commit.subsurface_position.x * parent->scale.w + subsurface->size.w + out->offset.x >= pointer->pos.x &&
+         subsurface->commit.subsurface_position.y * parent->scale.h + subsurface->size.h + out->offset.y >= pointer->pos.y)
+
```
Cloudef: useless space, also add { } as this if is multiline.

src/compositor/seat/pointer.c
```diff
@@ -240,6 +265,7 @@ wlc_pointer_button(struct wlc_pointer *pointer, uint32_t time, uint32_t button,
    except((seat = wl_container_of(pointer, seat, pointer)) && (compositor = wl_container_of(seat, compositor, seat)));
 
    // Special handling for popups
+
```
Cloudef: Useless whitespace

src/resources/types/surface.c
```diff
@@ -298,7 +300,29 @@ wlc_surface_release(struct wlc_surface *surface)
    struct wlc_surface_event ev = { .surface = surface, .type = WLC_SURFACE_EVENT_DESTROYED };
    wl_signal_emit(&wlc_system_signals()->surface, &ev);
 
-   wlc_handle_release(surface->view);
+   wlc_resource *sub;
+
+   if(!surface->parent) {
+      wlc_handle_release(surface->view);
+   } else {
+
```
Cloudef: Useless space

src/resources/types/surface.c
```diff
@@ -298,7 +300,29 @@ wlc_surface_release(struct wlc_surface *surface)
...
+         if(*sub == surface_id) {
+            chck_iter_pool_remove(&parent->subsurface_list, _I - 1);
+            break;
+         }
+      }
+   }
+
+   chck_iter_pool_for_each(&surface->subsurface_list, sub) {
+       wlc_surface_release(convert_from_wlc_resource(*sub, "surface"));
+   }
```
Cloudef: Make this, chck_iter_pool_for_each_call(&surface->subsurface_list, wlc_resource_release_ptr);
Avoid calling *_release for anything that is defined under resources/types/ as these types are tied to the lifetime of wayland resource.

src/compositor/seat/pointer.h
```diff
@@ -27,6 +27,16 @@ struct wlc_view;
 struct wlc_pointer_origin {
    double x, y;
 };
+// struct for keeping the focused surface
+// used because we cannot otherwise determine the position
+// of a subsurface without going all the way to the toplevel view
+// scale_w and scale_h are used only when the focused surface is the surface of the view
```
Cloudef: scale_w and scale_h isn't here anymore. wlc_focused_surface should make it already quite clear what this does IMO

src/compositor/view.c
```diff
@@ -18,6 +18,13 @@ configure_view(struct wlc_view *view, uint32_t edges, const struct wlc_geometry
 {
    assert(view && g);
 
+   struct wlc_surface *surface = convert_from_wlc_resource(view->surface, "surface");
+   /* we need scale only for visible surface */
+   if(surface && surface->size.w * surface->size.h > 0) {
+      surface->scale.w = (float)(g->size.w) / surface->size.w;
+      surface->scale.h = (float)(g->size.h) / surface->size.h;
```
Cloudef: Can the scale member be renamed to coordinate_scale or coordinate_transform perhaps, latter is usually used (it's usually matrix, but the type behind doesn't matter as long as we use the struct). Otherwise this can be easily confused that we are scaling surface, but we are actually just making input coordinates working with view which bounds aren't exact with the backing surface.

Cloudef:
This is starting to look quite good. I'm pretty sure the subsurfaces not appearing is due to the render loop not doing a cycle when it should (e.g. they only appear when application triggers new render cycle (moving pointer inside)). Though I may be wrong too.

Anyhow, I should give it spin too later.

src/compositor/output.c
rilix04: What do you mean? Sorry, I just don't get it.

ammen99: What do you mean? I'm sorry, I just don't get it.

ammen99:
My latest commit is from the account rilix04, this is because my sister was on the pc yesterday, her account was still logged in.

src/compositor/output.c
xerpi: I think Cloudef means adding an space between the if and the parenthesis, like this: if (!surface)

Cloudef: Yeah indeed, add space between if and parenthesis

ammen99:
Well, for the weston-subsurfaces bug: for me(I'm using orbment) I get different bug - the subsurfaces aren't placed correctly(their positions don't get commited) until I either resize the window(when nested compositor) or I switch the focus. I mean, when weston-subsurfaces start, it has some geometry. Orbment changes it to half the screen(I have a terminal opened before), the main surface gets changed, but the subsurfaces remain on their old positions. @Cloudef do you have idea what happens during these actions that causes weston-subsurfaces to commit the surfaces? This might show what's wrong.

ammen99:
I forgot to mention that this bug has nothing to do with synchronized or desynchronized subsurfaces. I tried directly committing position, which shows synchronization is irrelevant for this bug.

Cloudef:
What does the spec say about this. If the parent surface gets commit would the subsurfaces need to be commited as well by compositor?

src/compositor/seat/pointer.c
```diff
@@ -63,13 +63,43 @@ view_visible(struct wlc_view *view, uint32_t mask)
...
+find_surface_at_position_recursive(struct wlc_pointer *pointer, struct wlc_surface *parent, struct wlc_focused_surface *out) {
+
+   struct wlc_surface *surface = parent;
+   wlc_resource *sub;
+
+   chck_iter_pool_for_each(&parent->subsurface_list, sub) {
+      struct wlc_surface *subsurface = convert_from_wlc_resource(*sub, "surface");
+      if(!subsurface) {
+         wlc_log(WLC_LOG_ERROR, "Invalid subsurface id %" PRIuWLC " parent %" PRIuWLC,
+                 convert_to_wlc_resource(subsurface), convert_to_wlc_resource(parent));
```
Cloudef: Should continue; here.

Cloudef: You can actually crash there by opening gtk3 program and spawning tooltip. The subsurface isn't probably removed from the parent's iter pool.

ammen99:
According to the spec, if subsurface is synchronized, then when parent is commited subsurface is commited also. Maybe I should try to implement synchronization when I have more time and see what happens then.

src/compositor/seat/pointer.c
Cloudef: This log is also unneccessary here. I can add such log into resource.c and make it toggable with WLC_DEBUG (return warning each time NULL is returned from handle lookup).

ammen99:
I implemented subsurface synchronization. This fixes the problem with weston-subsurfaces - at least on my computer.

crondog:
Nice work. Weston-subsurfaces appears to be working correctly now. Don't forget you still need to clean up the stray printf and the if statements

ammen99:
I implemented subsurface_place_above/below requests, but do you know any program which uses these? I just have no way to test them.

Cloudef:
@ammen99 Not even the weston test programs? Could always write one yourself I guess. I think we could start merging this PR though, so we get later fixes / additions in smaller chunks.

ammen99:
I did your suggestions. In weston's tests there is one with such request - but it doesn't run(it shows a lot of errors), as do all other tests.

ammen99:
Is there something that still needs to be done or I should squash my commits into one to merge?

Cloudef:
Squash into one, and I will do final checkups that there is nothing obvious and merge it then.

Cloudef:
The changes in wlc_pointer_focus for transforming mouse coordinates (in case view geometry != client geometry) cause slight breakage. Pointer coordinates are wrong in programs. E.g. Launch weston-resizor, and try resize the window from bottom handle, you can't do it, but you can do it from bit more above.

I'm guessing from reading code, that this happens because the transformation does not handle correctly main surface (binded to view). Main surface coordinates are relative, and starts from 0,0. The calculation was changed in favor of subsurfaces which have different rules (?).

ammen99:
I fixed the bug. It turned out that the problem were maximized views, which now should be handled correctly. Please write if there are any other known bugs.

ammen99:
I didn't quite get what you mean. Moving the code to wlc_view_commit state definitely breaks pointer. The thing that I couldn't figure out before was why in configure_view() the parameter g is different from what I get when I call wlc_view_get_bounds()?

ammen99:
I hope this finally fixes the problem. It was simply because coordinate_scale must be updated not only when view is commited, but when surface->size changes(when commiting a buffer).

ammen99:
Maybe we should add a new function for surfaces. something like wlc_surface_update_coordinate_scale()?

Cloudef:
Yeah, that would be good idea. I think this should work inside view_commit too though, if not. There may be some bigger problem somewhere. I'll look into it too later.

Cloudef:
I'll look into why view_commit doesn't work for this now.

Cloudef:
I sent you a bit ugly PR, as I by instict updated it against latest wlc as well. That should make the coordinate transform updates bit saner. I think it's okay for merge after this. 

I'm wondering if the subsurfaces should be tight fitted with the main view when wlc forces it to requested geometry. It may be something to do in future, but for now that's not important. (It might rely on moving the wlc_view_get_bounds functionality to surface.c instead)

ammen99:
With these changes the coordinate scale doesn't get updated properly if I don't move the window. When  this happens though everything is OK. So we must either use surface_update_coordinate_transform or think of some other way to do this.

Cloudef:
It's probably because of the surface being created after view commit. And the precondition in view commit causes the function to be no-op if there was no changes to view state. If this is the case wlc_surface_set_parent should set the initial transform to new child surface.

ammen99:
I'm talking about the whole programs (try gedit for example). Maybe this should be done when surface size changes.

ammen99:
Do you have idea if there is already some function which gets triggered when surface size changes or some other related event?

Cloudef:
The view_commit_state is the function that applies new view size. For surfaces (not related to views) it's surface_commit (in surface.c).

ammen99:
So, after all, we must update coordinate_scale in each surface_commit, because we can't know when surface size changes. I will re-add the function as I see no other way. Tell me if I'm wrong.

ammen99:
I re-added the necessary code. Now everything works OK for me, maybe we should add check for if size has changed?

ammen99:
Is there something that still needs to be done or I should squash for merging?

Cloudef:
Squash, and I'll merge it.

Cloudef:
That did not go right I guess :P

ammen99:
Nope, it didn't. Sadly I cannot find a way to remove the merge using a rebase(it expands the commits)? Do you have idea why is this?

Cloudef:
Yeah, you should not ever need merge in PR's, just rebase on top of upstream should do it.

Cloudef:
Can you edit the commit message to be something more descriptive about this feature.

ammen99:
For example?

PR #123, opened 2016-02-17
Add link to golang bindings

mikkeloscar:
I guess it makes sense to add these here. I just updated the bindings to be completely compatible with the latest wlc version.


Cloudef:
Thanks, merged!

PR #124, opened 2016-02-21
Add link to rust bindings

SnirkImmington:
We have some Rust bindings around ~97% of wlc (excluding xkb bindings, libinput) and just updated to the latest version.

The link (in the commit) is https://github.com/Immington-Industries/rust-wlc.


Cloudef:
Thanks.

PR #131, opened 2016-03-03
Added environment variable AS_ROOT

Azarn:
Just thought that running from root could be useful.

### DIFF
--- a/src/wlc.c
+++ b/src/wlc.c
@@ -301,13 +301,14 @@ wlc_init(const struct wlc_interface *interface, int argc, char *argv[])
 
    unsetenv("TERM");
    const char *x11display = getenv("DISPLAY");
+   const char *as_root = getenv("AS_ROOT");
    bool privileged = false;
    const bool has_logind = wlc_logind_available();
 
    if (getuid() != geteuid() || getgid() != getegid()) {
       wlc_log(WLC_LOG_INFO, "Doing work on SUID/SGID side and dropping permissions");
       privileged = true;
-   } else if (getuid() == 0) {
+   } else if (!as_root && getuid() == 0) {
       die("Do not run wlc compositor as root");
    } else if (!x11display && !has_logind && access("/dev/input/event0", R_OK | W_OK) != 0) {
       die("Not running from X11 and no access to /dev/input/event0 or logind available");


Cloudef:
How is this useful? You are running lots of untrusted code and opening hole if AS_ROOT is set.

SirCmpwn:
I agree with this PR. It's not wlc's responsibility to babysit users, and there are often legitimate reasons to run it as root or not care about the security implications (i.e. working in a throwaway VM or something).

Cloudef:
What are these legimate reasons, give me some real life use cases so I can consider this?

Earnestly:
If this was added the correct way to do it would be to simply make wlc either warn or do nothing when running as root, not to add any kind of mechanism to request it.

I like tools which attempt to run as they're told and any errors and warnings should come from actual attempts to access resources any such which fail.  I dislike UID checks, such as in other tools like pacman which requires `fakeroot` hacks.

SirCmpwn:
Here's an example: I need to go from a fresh install of Arch to a working desktop ASAP to test something on a virtual machine. I don't care about the security of this machine, it's going to be destroyed in a few minutes anyway. The overhead of creating a user and setting things up for it is just an annoyance. (this scenario has happened to me when makepkg dropped support for being run as root, very annoying, especially because in that case I had to set up sudo and such)

Write software as if your users are smart enough to make their own informed decisions about it.

Earnestly:
*This is a tangent, but to be fair the [`nobody`](https://en.wikipedia.org/wiki/Nobody_(username)) user is available for this reason.*

SirCmpwn:
Getting an interactive session as nobody requires some setup on most distros, though. You'd get the same amount of setup necessary to just create a new user.

Cloudef:
Maybe just remove the root check if there is need for this. Env variable for this seems funny.

SirCmpwn:
Agreed.

nicman23:
you can do what makepkg used to do (still is?) in arch, were you could run as root if you run with `--as-root`

Earnestly:
@nicman23: Not since https://projects.archlinux.org/pacman.git/commit/?id=61ba5c961e

SirCmpwn:
I've always disagreed with that change, it's pretty far removed from the arch way imo.

SnirkImmington:
Maybe it should be included in the API itself so compositors can handle it differently? I'd prefer the `--as-root` flag but let the compositors handle how that works.

Earnestly:
This issue isn't about makepkg.

The proper solution here is to not have any root checking at all thus because pull request adds unnecessary code/complication, it likely won't be merged.

Try removing the `UID` checks instead, that would likely have more success.

Cloudef:
https://github.com/Cloudef/wlc/commit/46c44f763acb47ef08a3daafbca33ea909b00574

PR #135, opened 2016-03-07
Implement functions to get surface textures

### LINKED ISSUE #8: Provide animation API
Easings, timings, conflict handling and such coupled with custom drawing. Of course we want wobbly windows and desktop cubes. Having this API in wlc allows render loop to be efficient as possible.

ammen99:
This allows compositors to draw surfaces the way they want to,
so that they can also create custom animations.

New functions:
wlc_output_get_renderer()
wlc_surface_get_textures()
wlc_surface_get_subsurfaces()
wlc_get_subsurface_geometry()
wlc_surface_flush_frame_callbacks()

Maybe resolves #8 ?

### DIFF
--- a/include/wlc/wlc-render.h
+++ b/include/wlc/wlc-render.h
@@ -36,7 +36,6 @@ WLC_NONULL void wlc_pixels_write(enum wlc_pixel_format format, const struct wlc_
  */
 WLC_NONULL void wlc_pixels_read(enum wlc_pixel_format format, const struct wlc_geometry *geometry, struct wlc_geometry *out_geometry, void *out_data);
 
-
 /** Renders surface. */
 WLC_NONULL void wlc_surface_render(wlc_resource surface, const struct wlc_geometry *geometry);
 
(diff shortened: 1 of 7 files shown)


ammen99:
Updated

Cloudef:
When mask is 0, you are indicating the view is not visible. Wlc will optimize it out from render loop (similarly, if view is culled (behind something) the render hook will not be called).

Your suggestion would not work as the pre/post hooks per view are not called at all, only output post/pre (pre only if background is visible). This would need special casing, to tell wlc not to cull view with some option.

ammen99:
I meant not to set view mask to 0 at all, but use this additional flag as the way to tell wlc not to display the view on screen.

ammen99:
This is what I had on mind, if you can think of a better way, tell me.

Cloudef:
This will not work if the view is culled by wlc (Actually pre hook should not be called at all if mask == 0). I'd rather you give view or surface option in wlc-render for this specific purpose.

ammen99:
OK, will do this later.

ammen99:
This is what I came up with. Views with forced updates are added to output->visible even though they are not visible, but the flag view->state.forced_update shows that they shouldn't be rendered at all, only "updated"

ammen99:
What do you think of this? You were right, I forgot that frame callbacks can be pushed outside the render loop :)

Cloudef:
Yeah this is good.

ammen99:
Do you think this is ready for merge?

raycewest:
IIiya, can I PM/email you to discuss creating some animations?

ammen99:
Yes, sure. My email is iliyabo@hotmail.com

ammen99:
@Cloudef  I updated the commit.

Cloudef:
Thanks, merged.

Cloudef:
It's not technically correct. But the commits in wlc aren't anyways for now. (They sort of violate wayland protocol) I'm okay with that addition.

PR #141, opened 2016-03-28
Stop writepixels from crashing

crondog:
Not sure if this is a correct solution but it is here just for conversation

The crash is from https://github.com/SirCmpwn/sway/commit/7dc335bb46451ce984cb89bb7ed45ca932b2324d


Cloudef:
This seems very far fetched. But you might be on something, shm buffer attachs do set UNPACK_ROW_LENGTH_EXT which is never reset back to initial value (0).

crondog:
If i do GL_CALL(glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, 0)); at the end of shm_attach instead of the what i did in the pull it works

crondog:
Would that be the right place to do it? Or do you have somewhere better in mind to reset it?

SirCmpwn:
Something I can't believe I failed to mention until now is that window borders *do* render correctly the first time around, and continue to do so every frame until something changes.

Cloudef:
@crondog That would be the right place yes.

PR #143, opened 2016-03-29
Added handling of libinput gesture events

Hummer12007:
This PR adds handling of libinput gesture events with API for gesture.swipe and gesture.pinch.

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -122,6 +122,16 @@ enum wlc_touch_type {
    WLC_TOUCH_CANCEL,
 };
 
+/** Type in interface.gesture.swipe and interface.gesture.pinch funtions */
+enum wlc_gesture_type {
+   WLC_GESTURE_SWIPE_BEGIN,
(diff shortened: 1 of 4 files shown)


Cloudef:
This looks good so far. But is missing the actual handling of the touch signals in compositor/seat.c. I think I've implemented everything that can be passed to wl programs, but maybe the protocol has been updated since. If the programs can't catch these and are mainly meant for compositor to be read the bool return value doesn't really make sense as the input won't go to client anyways.

Hummer12007:
I'm working on it. Will push it in coming days.

tinchou:
@Hummer12007 any updates?? can I be of any help?

Hummer12007:
Oh, that reminds me. I'll take a look at during the week.

johan-bjareholt:
Any updates?

Hummer12007:
I forgot about it before, but it's definitely postponed now (as sway is probably migrating off wlc).

PR #145, opened 2016-03-31
Fix strage behavior for X11 child views

SirCmpwn:


SirCmpwn:
Fixes https://github.com/SirCmpwn/sway/issues/505

Issue #59, opened 2015-08-22
Create tty switch event handler


Luminarys:
Right now, outputs which are sleeping will be woken up again when the tty is switched. A handler for the tty switch event would prevent this.

Cloudef:
Output context change events were added that basically trigger this behaviour. It's not really TTY switch event per se, but indicates whenever the context for output was tear down'd and recreated. During creation the output would wake up.

https://github.com/Cloudef/wlc/commit/f21ec4055ff4c937e0c6968de75bbcbce5a47487

Cloudef:
If the above changes are enough for your needs, feel free close this issue.

PR #149, opened 2016-04-05
Implement new callbacks and several *fixes*

### LINKED ISSUE #137: tty switching when target tty  is more than 12
I noticed that the code which handles tty switching is in compositor/seat/seat.c Currently wlc uses hard-coded codes for the tty switching. But I think there is something you maybe don't want - the if below (at compositor/seat/seat.c around line 118) would switch to a different tty even when I press ctrl-alt-kp_1-9(the keypad) for example, which sends the user to non-existing tty(black screen with blinking cursor). Maybe you can change it so that it checks for `ev->key.code <= 71` which are only the keys f1-f12. I can send a pull request if you want me to do so.
Relevant code:
``` cpp                                                                                               
if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && ev->key.code >= 59 && ev->key.code <= 88) {                                         
   const int vt = (ev->key.code - 59) + 1;                                                                                                                       
   if (ev->key.state == WL_KEYBOARD_KEY_STATE_PRESSED && wlc_tty_get_vt() != vt) {                                                                               
      struct wlc_activate_event aev = { .active = false, .vt = vt };                                                                                             
      wl_signal_emit(&wlc_system_signals()->activate, &aev);                                                                                                     }
```

ammen99:
From what I saw, there is currently no signal which is emitted when the current tty is changed. When such a thing happens, the GL context is destroyed. This causes problems when the compositor has own GL state(textures, programs etc.) because these have to be recreated also. That's why we need the activated/deactivated signals. There is also a fix for switching to a tty >= 12. Lastly, why are alpha and depth buffers disabled? These are useful for creating a fade-in/out animation, as well as when rendering 3D animations.

Resolves #137

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -138,6 +138,14 @@ struct wlc_interface {
       WLC_NONULL void (*resolution)(wlc_handle output, const struct wlc_size *from, const struct wlc_size *to);
 
       struct {
+          /** Output context is created. This generally happens on startup and when current tty changes */
+          void (*created)(wlc_handle output);
+
+          /** Output context is destroyed */
+          void (*destroyed)(wlc_handle output);
+      } context;
+
+      struct {
          /** Pre render hook. */
          void (*pre)(wlc_handle output);
 
@@ -243,6 +251,12 @@ void wlc_set_output_render_pre_cb(void (*cb)(wlc_handle output));
 
 /** Output post render hook. */
 void wlc_set_output_render_post_cb(void (*cb)(wlc_handle output));
+
+/** Output context is created. This generally happens on startup and when current tty changes */
+void wlc_set_output_context_created_cb(void (*cb)(wlc_handle output));
+
+/** Output context was destroyed. */
+void wlc_set_output_context_destroyed_cb(void (*cb)(wlc_handle output));
 
 /** View was created. Return false if you want to destroy the view. (e.g. failed to allocate data related to view) */
 void wlc_set_view_created_cb(bool (*cb)(wlc_handle view));
--- a/src/compositor/compositor.c
+++ b/src/compositor/compositor.c
@@ -462,6 +462,8 @@ add_output(struct wlc_compositor *compositor, struct wlc_backend_surface *bsurfa
       return;
    }
 
+   output->state.created = true;
+
    if (!compositor->active.output)
       active_output(compositor, output);
 
--- a/src/compositor/output.c
+++ b/src/compositor/output.c
@@ -555,6 +555,9 @@ wlc_output_set_backend_surface(struct wlc_output *output, struct wlc_backend_sur
       }
    }
 
+   if (output->state.created)
+      WLC_INTERFACE_EMIT(output.context.destroyed, convert_to_wlc_handle(output));
+
    wlc_render_release(&output->render, &output->context);
    wlc_context_release(&output->context);
    wlc_backend_surface_release(&output->bsurface);
@@ -583,6 +586,9 @@ wlc_output_set_backend_surface(struct wlc_output *output, struct wlc_backend_sur
             wlc_surface_attach_to_output(s, output, wlc_surface_get_buffer(s));
          }
       }
+
+      if (output->state.created)
+         WLC_INTERFACE_EMIT(output.context.created, convert_to_wlc_handle(output));
 
       wlc_log(WLC_LOG_INFO, "Set new bsurface to output (%" PRIuWLC ")", convert_to_wlc_handle(output));
    } else {
--- a/src/compositor/seat/seat.c
+++ b/src/compositor/seat/seat.c
@@ -115,7 +115,13 @@ seat_handle_key(struct wlc_seat *seat, const struct wlc_input_event *ev)
 
    wlc_keyboard_update_modifiers(&seat->keyboard, ev->device);
 
-   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && ev->key.code >= 59 && ev->key.code <= 88) {
+   /* We use no mods to obtain keysym, because otherwise
+    * the ctrl-alt combo will change the resulting keysym
+    * into something different from KEY_F1 -> KEY_F12 */
+   struct wlc_modifiers mods = {0, 0};
+   uint32_t keysym = wlc_keyboard_get_keysym_for_key_ptr(&seat->keyboard, ev->key.code, &mods);
+
+   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && keysym >= XKB_KEY_F1 && keysym <= XKB_KEY_F12) {
       const int vt = (ev->key.code - 59) + 1;
       if (ev->key.state == WL_KEYBOARD_KEY_STATE_PRESSED && wlc_tty_get_vt() != vt) {
          struct wlc_activate_event aev = { .active = false, .vt = vt };
--- a/src/platform/context/egl.c
+++ b/src/platform/context/egl.c
@@ -177,7 +177,7 @@ create_context(struct wlc_backend_surface *bsurface)
             EGL_GREEN_SIZE, 1,
             EGL_BLUE_SIZE, 1,
             EGL_ALPHA_SIZE, 0,
-            EGL_DEPTH_SIZE, 0,
+            EGL_DEPTH_SIZE, 1,
             EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
             EGL_NONE
          }
--- a/src/compositor/output.h
+++ b/src/compositor/output.h
@@ -79,6 +79,7 @@ struct wlc_output {
       uint32_t frame_time;
       bool pending, scheduled, activity, sleeping;
       bool background_visible;
+      bool created;
    } state;
 
    struct {
--- a/src/wlc.c
+++ b/src/wlc.c
@@ -431,6 +431,18 @@ wlc_set_output_render_post_cb(void (*cb)(wlc_handle output))
 }
 
 WLC_API void
+wlc_set_output_context_created_cb(void (*cb)(wlc_handle output))
+{
+   wlc.interface.output.context.created = cb;
+}
+
+WLC_API void
+wlc_set_output_context_destroyed_cb(void (*cb)(wlc_handle output))
+{
+   wlc.interface.output.context.destroyed = cb;
+}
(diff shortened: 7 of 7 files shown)


src/compositor/seat/seat.c
```diff
@@ -115,7 +115,7 @@ seat_handle_key(struct wlc_seat *seat, const struct wlc_input_event *ev)
 
    wlc_keyboard_update_modifiers(&seat->keyboard, ev->device);
 
-   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && ev->key.code >= 59 && ev->key.code <= 88) {
+   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && ev->key.code >= 59 && ev->key.code <= 70) {
```
ammen99: You might want to have a look at this line. As you said, using plain integers is not a very good way to check this, but what other defines can we use(XKB_F1 definitely won't do it)

Cloudef:
You should not need alpha buffer unless compositor _itself_ has alpha. Depth buffer is only needed when you want depth testing (If you do 3d compositing this makes sense).

Thanks for the patch, I'll take a look.

src/compositor/seat/seat.c
Cloudef: The XKB symbols should work, but you can't use the .code as it means raw keycode and this may change depending on keyboard used. E.g. for me F12 == 96 (X11 historically adds 8 to keycodes, so 88 here). So this would break the current F range for me.

src/platform/context/egl.c
```diff
@@ -176,8 +176,8 @@ create_context(struct wlc_backend_surface *bsurface)
             EGL_RED_SIZE, 1,
             EGL_GREEN_SIZE, 1,
             EGL_BLUE_SIZE, 1,
-            EGL_ALPHA_SIZE, 0,
-            EGL_DEPTH_SIZE, 0,
+            EGL_ALPHA_SIZE, 1,
```
Cloudef: You should not need this for anything, unless you plan to render the compositor as transparent window over X11 or other compositor. This simply means that the color buffer for context will have alpha component, which does not make sense when rendering to TTY without anything behind.

src/compositor/compositor.c
```diff
@@ -257,6 +258,7 @@ deactivate_tty(struct wlc_compositor *compositor)
 
    compositor->state.tty = IDLE;
 
+   WLC_INTERFACE_EMIT(compositor.deactivated);
```
Cloudef: Minor nitpick, generally code does not hug {} or if clauses in wlc, you can just put this under the above expression .. state.tty = IDLE; ..

src/compositor/compositor.c
```diff
@@ -238,6 +238,7 @@ activate_tty(struct wlc_compositor *compositor)
    if (compositor->state.tty != ACTIVATING)
       return;
 
+   WLC_INTERFACE_EMIT(compositor.activated);
```
Cloudef: You probably want to do this after wlc_fd_activate.

Cloudef:
I'm not sure if relying in tty switch events is reliable for your use case. You'd rather want to get notified when output context is destroyed and when it's created again. Since the contexts are in outputs, and these outputs won't get immediately relased / created during tty switch. So something output context related callback may make more sense e.g. here: https://github.com/Cloudef/wlc/blob/master/src/compositor/output.c#L557

ammen99:
If I get what you mean, you suggest adding a signal when the bsurface of an output is changed, am I correct? Nevertheless, do you think that these can be useful or I should completely remove them from the PR?

Cloudef:
Yeah signal when bsurface is changed, if it's NULL it was removed completely (tty changed, compositor going down). I think people from sway requested me for TTY switch signals in past for something, they might be useful. Need to read history.

Cloudef:
Just adding for the bsurface signals that the order matters. You want to get signal for change before the destruction of current context, and signal after the new context was created for GL resource recreation to work right. (Sucks that we need this in first place though :()

ammen99:
Related issue is #59

src/compositor/seat/seat.c
```diff
@@ -114,8 +114,10 @@ seat_handle_key(struct wlc_seat *seat, const struct wlc_input_event *ev)
       return;
 
    wlc_keyboard_update_modifiers(&seat->keyboard, ev->device);
+   struct wlc_modifiers mods = {0, 0};
+   uint32_t keysym = wlc_keyboard_get_keysym_for_key_ptr(&seat->keyboard, ev->key.code, &mods);
 
-   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && ev->key.code >= 59 && ev->key.code <= 88) {
+   if (seat->keyboard.modifiers.mods == (WLC_BIT_MOD_CTRL | WLC_BIT_MOD_ALT) && keysym >= XKB_KEY_F1 && keysym <= XKB_KEY_F12) {
```
ammen99: You might want to test this, for me it works :)

Cloudef: As long as the keysym constants increase linearly from F1 -> F12 (which I hope they do) This should be fine.

src/platform/context/egl.c
```diff
@@ -176,7 +176,7 @@ create_context(struct wlc_backend_surface *bsurface)
             EGL_RED_SIZE, 1,
             EGL_GREEN_SIZE, 1,
             EGL_BLUE_SIZE, 1,
-            EGL_ALPHA_SIZE, 1,
+            EGL_ALPHA_SIZE, 0,
```
Cloudef: Whoops, not related to this commit :)

src/compositor/seat/seat.c
```diff
@@ -114,8 +114,10 @@ seat_handle_key(struct wlc_seat *seat, const struct wlc_input_event *ev)
       return;
 
    wlc_keyboard_update_modifiers(&seat->keyboard, ev->device);
+   struct wlc_modifiers mods = {0, 0};
```
Cloudef: I assume this is used intentionally instead of seat->keyboard.modifiers to avoid translation caused by CTRL + ALT to some other key symbol?

Cloudef: Use {0}; instead btw.

src/compositor/seat/seat.c
ammen99: Otherwise, if you prefer it, we can create a simple function to return the appropriate tty number or  -1 if the key is not in this range.

src/compositor/seat/seat.c
ammen99: Yes. If I use keyboard->modifiers, we get the wrong keysym, because it includes the ctrl+alt too.

Cloudef: Ok, may make comment about it. That's what I thought though :)

src/compositor/seat/seat.c
Cloudef: Looking at the keysym constants they seem to rise linearly, I'm fine with this. I doubt that they'll change. There may be problem if somebody provides XKB implementation that has different constant that for some reason doesn't work this way. But if that ever happens lets worry only then. This is more straightforward so I like it.

src/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ foreach (src ${sources})
    set_source_files_properties(${src} PROPERTIES COMPILE_FLAGS -DWLC_FILE=\\\"${src}\\\")
 endforeach ()
 
-add_compile_options(-fvisibility=hidden)
+add_compile_options(-g)
```
Cloudef: Hm?

src/compositor/output.c
```diff
@@ -539,7 +539,7 @@ wlc_output_set_backend_surface(struct wlc_output *output, struct wlc_backend_sur
    if (output->state.pending) {
       wlc_log(WLC_LOG_INFO, "Pending bsurface set for output (%" PRIuWLC ")", convert_to_wlc_handle(output));
       if (bsurface) {
-         memcpy(&output->task.bsurface, bsurface, sizeof(output->task.bsurface));
+        memcpy(&output->task.bsurface, bsurface, sizeof(output->task.bsurface));
```
Cloudef: Accident I guess.

include/wlc/wlc.h
```diff
@@ -138,6 +137,14 @@ struct wlc_interface {
       WLC_NONULL void (*resolution)(wlc_handle output, const struct wlc_size *from, const struct wlc_size *to);
 
       struct {
+          /** Output context is created. This generally happens on startup and when current tty changes */
+          void (*created)(wlc_handle output);
+
+          /** Output context is destroyed */
+          void (*destroyed)(wlc_handle output);
+      } ctx;
```
Cloudef: In general wlc hasn't use shortened words in the interface, so I would just make this "context"

Cloudef:
Looks good. There are 3 different changes in this PR, so when you squash. Keep TTY key change, context interface additions, and EGL_ALPHA_SIZE change in own commits.

So you would end up with 3 commits after squash.

ammen99:
Ready

Cloudef:
Looks good. I'll give it a spin later today. There is one issue I foresee regarding the context callbacks that they may come before output.created. We probably need to do same thing as view's do that they hold `bool created` in their datastructure, and don't pass other callbacks until the created callback was signaled.

https://github.com/ammen99/wlc/blob/master/src/compositor/compositor.c#L460

ammen99:
I just tested and what you said is correct. Context callback is called before output.create. Do you think that it would be sane to ignore the first context creation callback, as it happens exactly when output is created?

src/compositor/output.c
```diff
@@ -1027,6 +1029,7 @@ wlc_output(struct wlc_output *output)
 
    output->active.mode = UINT_MAX;
    output->state.ims = 41;
+   output->state.created = false;
```
Cloudef: Don't need this as the output struct is zero initialized.

src/compositor/compositor.c
```diff
@@ -462,6 +462,7 @@ add_output(struct wlc_compositor *compositor, struct wlc_backend_surface *bsurfa
       return;
    }
 
+   output->state.created = true;
```
Cloudef: space out from the if block and it's fine :P

Cloudef:
What you did in the new commit is more reliable way IMO.

Cloudef:
:+1: I'll test the tty switches myself as well later today and merge it then.

PR #150, opened 2016-04-06
Make X11 backend and Xwayland support optional

cl91:
This makes the X11 backend and Xwayland support optional at compile time --- if relevant X11 and XCB packages are found, then they are enabled. If not, they are automatically disabled.

This would be useful for systems that don't need X11 and don't want any X11 dependencies (for instance, embedded systems). I'm currently building a Wayland-only Gentoo system (that is, no X11 dependencies at all), and have successfully tested sway with this commit (In fact, I'm writing this under sway and midori right now. All built without any X11 dependencies (libX11 and friends).


Cloudef:
https://github.com/Cloudef/wlc/issues/47 The correct way is to do this issue.

I don't generally like how many #ifdef paths this patch introduces, especially in places that are not directly related to X11 or Xwayland.

If you can reduce that clutter by no-op:ing datastructures and their access functions instead of disabling code, I would accept this.

E.g. Instead of doing 
```
+#ifdef ENABLE_XWAYLAND
     wlc_xwm_release(&compositor->xwm);
 +#endif
```

Do the #ifdef inside the wlc_xwm_release. The xwm basically would be no-op datastructure when it's not compiled in.

cl91:
I have refactored the code to remove the #ifdef's and added no-ops and dummy data structures for the case where xwayland is disabled. It should look much cleaner now.

src/compositor/view.h
```diff
@@ -59,6 +59,10 @@ struct wlc_view {
    } state;
 };
 
+static inline bool is_x11_view(struct wlc_view *view) {
```
Cloudef: Put storage class and return value in own line e.g:
```
static inline bool
is_x11_view(...)
{
   ...
}
```

This is the style wlc uses.

include/wlc/defines.h
```diff
@@ -12,11 +12,13 @@ extern "C" {
 #  define WLC_NONULLV(...) __attribute__((nonnull(__VA_ARGS__)))
 #  define WLC_PURE __attribute__((pure))
 #  define WLC_CONST __attribute__((const))
+#  define WLC_UNUSED __attribute__((unused))
```
Cloudef: Instead of __attribute__((unused)) wlc has used to (void)foobar; to ignore "not used" warnings.

src/xwayland/xwm.h
```diff
@@ -36,4 +53,43 @@ WLC_NONULL void wlc_x11_window_close(struct wlc_x11_window *win);
...
+
+WLC_NONULL static inline void wlc_x11_set_window_hidden(struct wlc_x11_window *w WLC_UNUSED, bool hidden WLC_UNUSED)
+{
+}
+
+WLC_NONULL static inline void wlc_x11_window_set_surface_format(struct wlc_surface *surface WLC_UNUSED, struct wlc_x11_window *win WLC_UNUSED) {}
+WLC_NONULL static inline void wlc_x11_window_configure(struct wlc_x11_window *win WLC_UNUSED, const struct wlc_geometry *g WLC_UNUSED) {}
+WLC_NONULL static inline void wlc_x11_window_set_state(struct wlc_x11_window *win WLC_UNUSED, enum wlc_view_state_bit state WLC_UNUSED, bool toggle WLC_UNUSED) {}
+
+WLC_NONULL static inline bool wlc_x11_window_set_active(struct wlc_x11_window *win WLC_UNUSED, bool active WLC_UNUSED)
```
Cloudef: Remove WLC_UNUSED and put `(void)win; (void)active;` inside function body.

Cloudef:
Yup it's much better contained now. Fix the new comments and then it should be ok for squash and merge.

cl91:
I have fixed the style issues with the new commits now.

Cloudef:
Thanks, it looks good. Squash and I will merge it :)

cl91:
Done.

Cloudef:
Thanks, will smoke the builds later today.

PR #155, opened 2016-04-22
Initial documentation

Earnestly:
The `man3` stuff should be okay to go, I'm not sure how you feel about having a second `CONTRIBUTING.rst` and the WIP section 7 manual could be left out until more complete.

This will need some cmake work where the manuals are installed to (using GNU's terminology) `datarootdir/mandir/man(3,7)/` which is usually `/share/man/man(3,7)`


Cloudef:
Looks good, :+1: for the contribution guide. Will you be adding more commits still?

Earnestly:
Yeah, I've got a few more queued.

Drakulix:
`The parameter time represents a timestamp of when the mouse event occured. It is a magical number and cannot be relied on.`
I was looking for some information regarding this timestamp and THIS is the best I've found. Does this mean the argument is effectively useless or may I kindly ask why it is *magical* and what I can learn from it. :smile:

Cloudef:
I think it's magical in sense that it depends on what backend puts on it and I did not even standardize the unit.. :) For the common case though (running from TTY) that timestamp comes from libinput and will correspond to whatever libinput gives. But the documentation is correct in this case, that the argument is badly specified.

Drakulix:
@Cloudef Alright, thanks! So I guess the only reasonable thing you may do with this value is comparing it to other events?

Cloudef:
Yeah, it's mainly provided so the callback can give you events from the past.

Fale:
any news on those man pages?

Earnestly:
I haven't been working on many lately

Timidger:
What is left to document? I can look into it because I'd rather not have to duplicate documentation for [rust-wlc](http://immington-industries.github.io/rustwlc/index.html)

Earnestly:
@Timidger https://github.com/Cloudef/wlc/issues/156

telent:
These docs are awesome, I wish I'd known about this branch when I started looking at wlc.  I won't ask "can it be merged" because free software, your priorities are your own, etc etc, but as a consumer of the software I definitely think it's worth merging as-is irrespective of any further updates that may or may not be forthcoming.

PR #159, opened 2016-04-30
Coding style: use local include

abique:


Earnestly:
Why?

abique:
Because #include "" uses relative path.
While #include <> uses search path.

abique:
And by the way you should also include standard headers first, not yours. That way your marcos will not mess with standard headers.

Earnestly:
Oh I'm not disagreeing but you need explain why in the message body.  You also need to explain why using relative paths is better.  I would hate to see crap like `#include "../foo.h"` but that doesn't seem to be what you're advocating here.

abique:
Do it as it please.

Earnestly:
@abique I'm not the owner of this repository incase you didn't notice, however I think asking for a little justification is a reasonable request.

There's really no need to be callow.

PR #163, opened 2016-05-01
Fix incorrect calculation of refresh rate in Hz

b52:
Neither the input value nor the output value are mHz, thus the comment is somewhat of. Additionally, the value contained in `vrefresh` is already in Hz, therefore we don't have to do anything.


Cloudef:
Wayland protocol specifies the refresh rate in mHz, you either need to convert it there, or when feeding it to clients. (X11 backend does this conversion as well).

PR #168, opened 2016-05-20
CMake: Add an explicit option for X11 support

sardemff7:
Not sure if more magic is wanted.

If you want another option name, please tell.


lib/chck
```diff
@@ -1 +1 @@
-Subproject commit ab3d82e3dab626682844f5000c2a5f9eecf5e827
+Subproject commit 27c9ecfc0639f225d0b1dbdbd1874a58b18d0949
```
Cloudef: This is probably mistake?

PR #169, opened 2016-05-27
xdg_surface to negative value origin

athrungithub:
fix geometry origin to positive value

### DIFF
--- a/src/compositor/view.c
+++ b/src/compositor/view.c
@@ -203,8 +203,8 @@ wlc_view_get_bounds(struct wlc_view *view, struct wlc_geometry *out_bounds, stru
 
    if (view->xdg_surface && view->surface_commit.visible.size.w > 0 && view->surface_commit.visible.size.h > 0) {
       // xdg-surface client that draws drop shadows or other stuff.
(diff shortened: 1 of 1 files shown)


athrungithub:
I sorry, no use it for now, introduces new bug in app with window decoration gtk (sure) same with weston-apps with window toolkit.
Error message:
compositor/seat/data.c: 150 @ wlc_data_device_manager_offer(): Tried to retrieve handle of wrong type (surface != data-source)

Cloudef:
That message is not related to this pr, it's probably just bug in the clipboard support of wlc.
As for this PR, I'm not sure what issue does this fix. What I see this is potentially breaking tiling of windows with shadow as decorations and such.

athrungithub:
i use dmenu with origin and width ej dmenu -l 20 -x 100 -y 200 -w 250 .
i compile bemenu with xdg_surface for set geometry, for same purpose but, need introduce origin x, y in negative value.  ej bemenu -l 20 -x -100 -y -200 -w 250

Cloudef:
Clients cant set their x, y positions without extra protocol extension. (Which is frowned upon on wayland). Orbment just hardcodes bemenu at the top.

Issue #57, opened 2015-08-22
Handle surface scale parameter


Cloudef:
Needed for hidpi clients and scaling non hidpi clients.

Cloudef:
Note that surface scale is integral, this is for very good reasons which you can read on wayland mailing lists. Non-integral scaling can be still done by setting output resolution with wlc_output_set_resolution.

JonathanReeve:
:+1: for this all the way. This is what's preventing me, and everyone else with a modern display, from using SirCmpwn/sway.

SirCmpwn:
@JonathanReeve it helps as a workaround to set your display's resolution to half of the actual resolution in your sway config. That's what I do for my macbook.

harrisonrodgers:
Hi. Could the [scale](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output-event-scale) parameter of [wl_output](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output) be what we want? 
> A scale larger than 1 means that the compositor will automatically scale surface buffers by this amount when rendering. This is used for very high resolution displays where applications rendering at the native resolution would be too small to be legible.

SirCmpwn:
That's partially related, but not the whole picture.

harrisonrodgers:
Ok. Adding `scale=2` into an `[output]` section of the Weston config seems to scale everything as expected for a HiDPI monitor. Maybe we can look at the Weston code to see how exactly they do it.

Cloudef:
In order to do this, you need both scale parameter in wl_output and scale parameter in wl_surface. If surface scale == output scale, you don't need to do anything, the client is already rendering in native resolution. If surface scale > output scale, you need to downscale the surface. If surface scale < output scale, you need to scale the surface up.

The scaling should be done during render phase in gles2.c, it won't affect anything else but visual output. The visible bounds probably need to be multiplied / divided accordingly.

enckse:
If/when hidpi is available I'd be more than willing to try it (sway + wlc) via Arch and/or AUR.

Mildly unrelated my use case for hidpi is a dell 7370 where I would like to get my 2 external monitors + laptop screen all working (in sway) - I know configuration for multiple monitors may not _entirely_ fall on wlc but if there are good resources for that and someone could point me to the docs that would be fantastic (again since I'll have 1 hidpi display and 2 non-hidpi displays)

Cloudef:
[...]

PR #173, opened 2016-06-21
handle set_buffer_scale and output_scale

ss1h2a3tw:
Add set output scale, get output scale,get output scaled size api
with @david50407 's help
for fixing https://github.com/Cloudef/wlc/issues/57
I'm not sure about whether the pointer handling is right, I don't know how to check the recursive part.
Wish to make some progress to HiDPI support XD.

### DIFF
--- a/example/example.c
+++ b/example/example.c
@@ -4,6 +4,8 @@
 #include <chck/math/math.h>
 #include <linux/input.h>
 
+#define SCALE 2
+
 static struct {
    struct {
       wlc_handle view;
@@ -72,18 +74,19 @@ relayout(wlc_handle output)
    // very simple layout function
    // you probably don't want to layout certain type of windows in wm
 
-   const struct wlc_size *r;
-   if (!(r = wlc_output_get_resolution(output)))
+   // We are going to use scaled size here,
+   // that we don't have to divide the resolution for scaling
+   struct wlc_size s;
+   if (!(wlc_output_get_scaled_size(output, &s)))
       return;
 
    size_t memb;
    const wlc_handle *views = wlc_output_get_views(output, &memb);
-
    bool toggle = false;
    uint32_t y = 0;
-   uint32_t w = r->w / 2, h = r->h / chck_maxu32((1 + memb) / 2, 1);
+   uint32_t w = s.w / 2, h = s.h / chck_maxu32((1 + memb) / 2, 1);
    for (size_t i = 0; i < memb; ++i) {
-      struct wlc_geometry g = { { (toggle ? w : 0), y }, { (!toggle && i == memb - 1 ? r->w : w), h } };
+      struct wlc_geometry g = { { (toggle ? w : 0), y }, { (!toggle && i == memb - 1 ? s.w : w), h } };
       wlc_view_set_geometry(views[i], 0, &g);
       y = y + (!(toggle = !toggle) ? h : 0);
    }
@@ -92,6 +95,7 @@ relayout(wlc_handle output)
 static void
 output_resolution(wlc_handle output, const struct wlc_size *from, const struct wlc_size *to)
 {
+   wlc_output_set_scale(output, SCALE);
    (void)from, (void)to;
    relayout(output);
 }
@@ -254,7 +258,6 @@ int
 main(void)
 {
    wlc_log_set_handler(cb_log);
-
    wlc_set_output_resolution_cb(output_resolution);
    wlc_set_view_created_cb(view_created);
    wlc_set_view_destroyed_cb(view_destroyed);
--- a/example/example.c
+++ b/example/example.c
@@ -4,6 +4,8 @@
 #include <chck/math/math.h>
 #include <linux/input.h>
 
+/* We are using output scale = 2 as example
+ * So everything will be double sized than normal*/
 #define SCALE 2
 
 static struct {
@@ -258,6 +260,7 @@ int
 main(void)
 {
    wlc_log_set_handler(cb_log);
+
    wlc_set_output_resolution_cb(output_resolution);
    wlc_set_view_created_cb(view_created);
    wlc_set_view_destroyed_cb(view_destroyed);
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -284,6 +284,15 @@ const struct wlc_size* wlc_output_get_resolution(wlc_handle output);
 /** Set resolution. */
 WLC_NONULL void wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution);
 
+/** Get scale factor. */
+int32_t wlc_output_get_scale(wlc_handle output);
+
+/** Set scale factor. */
+void wlc_output_set_scale(wlc_handle output, int32_t scale);
+
+/** Get scaled size. */
+bool wlc_output_get_scaled_size(wlc_handle output, struct wlc_size *out_size);
+
 /** Get current visibility bitmask. */
 uint32_t wlc_output_get_mask(wlc_handle output);
 
--- a/src/compositor/output.c
+++ b/src/compositor/output.c
@@ -868,10 +868,50 @@ wlc_output_get_resolution(wlc_handle output)
    return get(convert_from_wlc_handle(output, "output"), offsetof(struct wlc_output, resolution));
 }
 
+bool
+wlc_output_set_scale_ptr(struct wlc_output *output, int32_t scale)
+{
+   if (!output)
+      return false;
+
+   assert(scale);
+   wlc_log(WLC_LOG_WARN, "Set scale of output %" PRIuWLC "to %u", convert_to_wlc_handle(output), scale);
+   output->information.scale=scale;
+   return true;
+}
 WLC_API void
 wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution)
 {
    wlc_output_set_resolution_ptr(convert_from_wlc_handle(output, "output"), resolution);
+}
+
+WLC_API void
+wlc_output_set_scale(wlc_handle output, int32_t scale)
+{
+   wlc_output_set_scale_ptr(convert_from_wlc_handle(output, "output"), scale);
+}
+WLC_API int32_t
+wlc_output_get_scale(wlc_handle output)
+{
+   struct wlc_output *o = convert_from_wlc_handle(output, "output");
+   assert(o);
+   return o->information.scale;
+}
+
+WLC_API bool
+wlc_output_get_scaled_size(wlc_handle output, struct wlc_size *out_size)
+{
+   if (!out_size)
+      return false;
+
+   const struct wlc_size *resolution;
+   if (!(resolution = wlc_output_get_resolution(output)))
+      return false;
+
+   out_size->w = resolution->w / wlc_output_get_scale(output);
+   out_size->h = resolution->h / wlc_output_get_scale(output);
+
+   return true;
 }
 
 WLC_API bool
--- a/src/compositor/output.c
+++ b/src/compositor/output.c
@@ -868,6 +868,12 @@ wlc_output_get_resolution(wlc_handle output)
    return get(convert_from_wlc_handle(output, "output"), offsetof(struct wlc_output, resolution));
 }
 
+WLC_API void
+wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution)
+{
+   wlc_output_set_resolution_ptr(convert_from_wlc_handle(output, "output"), resolution);
+}
+
 bool
 wlc_output_set_scale_ptr(struct wlc_output *output, int32_t scale)
 {
@@ -875,14 +881,8 @@ wlc_output_set_scale_ptr(struct wlc_output *output, int32_t scale)
       return false;
 
    assert(scale);
-   wlc_log(WLC_LOG_WARN, "Set scale of output %" PRIuWLC "to %u", convert_to_wlc_handle(output), scale);
    output->information.scale=scale;
    return true;
-}
-WLC_API void
-wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution)
-{
-   wlc_output_set_resolution_ptr(convert_from_wlc_handle(output, "output"), resolution);
 }
 
 WLC_API void
@@ -890,6 +890,7 @@ wlc_output_set_scale(wlc_handle output, int32_t scale)
 {
    wlc_output_set_scale_ptr(convert_from_wlc_handle(output, "output"), scale);
 }
+
 WLC_API int32_t
 wlc_output_get_scale(wlc_handle output)
 {
--- a/src/compositor/seat/pointer.c
+++ b/src/compositor/seat/pointer.c
@@ -139,10 +139,10 @@ pointer_paint(struct wlc_pointer *pointer, struct wlc_output *output)
 
    if (!pointer || output != active_output(pointer))
       return;
-
+   const int32_t scale=output->information.scale;
    const struct wlc_point pos = {
-      chck_clamp(pointer->pos.x, 0, output->resolution.w),
-      chck_clamp(pointer->pos.y, 0, output->resolution.h)
+      chck_clamp(pointer->pos.x*scale, 0, output->resolution.w),
+      chck_clamp(pointer->pos.y*scale, 0, output->resolution.h)
    };
 
    struct wlc_view *view = convert_from_wlc_handle(pointer->focused.view, "view");
@@ -214,6 +214,9 @@ focus_view(struct wlc_pointer *pointer, struct wlc_surface *surf, wlc_handle old
    if (!surf || !(surface = convert_to_wl_resource(surf, "surface")))
       return;
 
+   uint32_t scale = surf->pending.scale;
+   if(scale==0)scale=1;
+
    struct wl_client *client = wl_resource_get_client(surface);
    wlc_resource *r;
    chck_pool_for_each(&pointer->resources.pool, r) {
@@ -225,7 +228,7 @@ focus_view(struct wlc_pointer *pointer, struct wlc_surface *surf, wlc_handle old
          wlc_log(WLC_LOG_WARN, "Failed to push focused pointer resource to pool (out of memory?)");
 
       uint32_t serial = wl_display_next_serial(wlc_display());
-      wl_pointer_send_enter(wr, serial, surface, wl_fixed_from_double(pos->x), wl_fixed_from_double(pos->y));
+      wl_pointer_send_enter(wr, serial, surface, wl_fixed_from_double(pos->x/scale), wl_fixed_from_double(pos->y/scale));
    }
 
    pointer->focused.surface.id = convert_to_wlc_resource(surf);
@@ -330,8 +333,13 @@ wlc_pointer_motion(struct wlc_pointer *pointer, uint32_t time, bool pass)
    surface_under_pointer(pointer, output, &focused);
    pointer->focused.surface.offset = focused.offset;
 
+   struct wlc_surface *surface = convert_from_wlc_resource(focused.id, "surface");
+
    if (pass)
-      wlc_pointer_focus(pointer, convert_from_wlc_resource(focused.id, "surface"), &d);
+      wlc_pointer_focus(pointer, surface, &d);
+
+   uint32_t scale = surface->pending.scale;
+   if(scale==0)scale=1;
 
    wlc_output_schedule_repaint(output);
 
@@ -344,7 +352,7 @@ wlc_pointer_motion(struct wlc_pointer *pointer, uint32_t time, bool pass)
       if (!(wr = wl_resource_from_wlc_resource(*r, "pointer")))
          continue;
 
-      wl_pointer_send_motion(wr, time, wl_fixed_from_double(d.x), wl_fixed_from_double(d.y));
+      wl_pointer_send_motion(wr, time, wl_fixed_from_double(d.x/scale), wl_fixed_from_double(d.y/scale));
    }
 }
 
--- a/src/compositor/seat/pointer.c
+++ b/src/compositor/seat/pointer.c
@@ -214,8 +214,7 @@ focus_view(struct wlc_pointer *pointer, struct wlc_surface *surf, wlc_handle old
    if (!surf || !(surface = convert_to_wl_resource(surf, "surface")))
       return;
 
-   uint32_t scale = surf->pending.scale;
-   if(scale==0)scale=1;
+   const int32_t scale = surf->commit.scale;
 
    struct wl_client *client = wl_resource_get_client(surface);
    wlc_resource *r;
@@ -338,8 +337,7 @@ wlc_pointer_motion(struct wlc_pointer *pointer, uint32_t time, bool pass)
    if (pass)
       wlc_pointer_focus(pointer, surface, &d);
 
-   uint32_t scale = surface->pending.scale;
-   if(scale==0)scale=1;
+   const int32_t scale = surface->commit.scale;
 
    wlc_output_schedule_repaint(output);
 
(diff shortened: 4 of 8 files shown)


example/example.c
```diff
@@ -4,6 +4,8 @@
 #include <chck/math/math.h>
 #include <linux/input.h>
 
+#define SCALE 2
```
jserv: Why did you hardcode `2` here?

example/example.c
```diff
@@ -254,7 +258,6 @@ int
 main(void)
 {
    wlc_log_set_handler(cb_log);
-
```
jserv: Don't change the original coding style.

src/compositor/output.c
```diff
@@ -868,12 +868,52 @@ wlc_output_get_resolution(wlc_handle output)
...
 }
 
+bool
+wlc_output_set_scale_ptr(struct wlc_output *output, int32_t scale)
+{
+   if (!output)
+      return false;
+
+   assert(scale);
+   wlc_log(WLC_LOG_WARN, "Set scale of output %" PRIuWLC "to %u", convert_to_wlc_handle(output), scale);
```
jserv: There is no flags following character `%` in the format string.

src/compositor/output.c
```diff
@@ -868,12 +868,52 @@ wlc_output_get_resolution(wlc_handle output)
...
 wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution)
 {
    wlc_output_set_resolution_ptr(convert_from_wlc_handle(output, "output"), resolution);
 }
 
+WLC_API void
+wlc_output_set_scale(wlc_handle output, int32_t scale)
+{
+   wlc_output_set_scale_ptr(convert_from_wlc_handle(output, "output"), scale);
+}
```
jserv: Append the space line to separate each function.

src/compositor/seat/pointer.c
```diff
@@ -214,6 +214,9 @@ focus_view(struct wlc_pointer *pointer, struct wlc_surface *surf, wlc_handle old
    if (!surf || !(surface = convert_to_wl_resource(surf, "surface")))
       return;
 
+   uint32_t scale = surf->pending.scale;
+   if(scale==0)scale=1;
```
jserv: Follow the consistent coding style.

src/platform/render/gles2.c
```diff
@@ -734,6 +734,19 @@ view_paint(struct ctx *context, struct wlc_view *view)
 
    struct wlc_geometry geometry;
    wlc_view_get_bounds(view, &geometry, &settings.visible);
+   wlc_handle output = wlc_view_get_output(convert_to_wlc_handle(view));
+
+   int32_t scale=wlc_output_get_scale(output);
+   //wlc_log(WLC_LOG_WARN, "origin x %d y %d size x %u y %u  scale %u",geometry.origin.x,geometry.origin.y,geometry.size.w,geometry.size.h,scale);
```
jserv: Don't leave the dead code.

src/platform/render/gles2.c
```diff
@@ -734,6 +734,19 @@ view_paint(struct ctx *context, struct wlc_view *view)
...
+   geometry.origin.x*=scale;
```
jserv: Follow the consistent coding style.

include/wlc/wlc.h
```diff
@@ -284,6 +284,15 @@ const struct wlc_size* wlc_output_get_resolution(wlc_handle output);
...
 WLC_NONULL void wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution);
 
+/** Get scale factor. */
+int32_t wlc_output_get_scale(wlc_handle output);
+
+/** Set scale factor. */
+void wlc_output_set_scale(wlc_handle output, int32_t scale);
+
+/** Get scaled size. */
+bool wlc_output_get_scaled_size(wlc_handle output, struct wlc_size *out_size);
```
jserv: Is function `wlc_output_get_scaled_size` necessary? Can we simply calculate demanding size each time?

example/example.c
ss1h2a3tw: This is a example of a compositor working on scale 2.

jserv: You should specify it to reflect its real functionality.

include/wlc/wlc.h
ss1h2a3tw: It is not necessary, but useful. Because developer may meet the situation that there are multiple monitors with different scale. With this function, developer can directly get the global coordinate of the compositor.

src/platform/render/gles2.c
ss1h2a3tw: Sorry, I miss this line when I cleaning it up.

example/example.c
ss1h2a3tw: OK

jserv:
@ss1h2a3tw : You should squash the coding style fixes into single commit via `git rebase -i`.

include/wlc/wlc.h
david50407: @jserv Yes, it's not necessary, but if we considered that compositor use the global coordinate which's dimension is same as the surface coordinate. 
That would be useful that developer of the compositor can just get the result from wlc without handle this my themselves.

src/compositor/output.c
```diff
@@ -868,12 +868,52 @@ wlc_output_get_resolution(wlc_handle output)
...
+WLC_API int32_t
+wlc_output_get_scale(wlc_handle output)
+{
+   struct wlc_output *o = convert_from_wlc_handle(output, "output");
+   assert(o);
+   return o->information.scale;
+}
+
+WLC_API bool
```
jserv: Why does it return boolean here?

ss1h2a3tw: @david50407

david50407: @jserv check if the operation is succeed. Fails may caused by NULL out param or cannot get the resolution of the output (may a invalid output).

jserv: I think it would be straightforward if a compositor takes care of scaling information and makes precise decision itself.

david50407: wlc is for helping to building compositor, wlc could be a useful interface layer for developers to make a compositor easily and quickly, but if we wants to get the original resolution, we still can use `wlc_output_get_resolution`.
This is just a helper function.

david50407: And for some scaling tricks, we wish that wlc can handle non-integer scaling, and the fake coordinate will be calculated in wlc, so that developers can make the compositor in the global  coordinate world directly.

src/compositor/output.c
Cloudef: Hmm doesn't it make sense to have output_set_scale in public API? Normally I would assume the backend handles the scales.

include/wlc/wlc.h
Cloudef: Right now wlc doesn't have concept of global coordinates (and may not have, as there is no standard for monitor placement)

src/compositor/output.c
Cloudef: non-integral scaling can be achieved by combination of scale and wlc_output_set_resolution. Non integral scaling for surfaces is not supported by wayland protocol. Also supporting that would be really PITA when you consider buffer sizes could be non integral. Thus it's better to set nearest integral scale to output and then compensate with rendering resolution (not neccessarily modeline).

src/compositor/seat/pointer.c
```diff
@@ -139,10 +139,10 @@ pointer_paint(struct wlc_pointer *pointer, struct wlc_output *output)
 
    if (!pointer || output != active_output(pointer))
       return;
-
+   const int32_t scale=output->information.scale;
```
Cloudef: whitespace around operators.

src/compositor/output.c
```diff
@@ -874,6 +874,47 @@ wlc_output_set_resolution(wlc_handle output, const struct wlc_size *resolution)
...
+wlc_output_get_scaled_size(wlc_handle output, struct wlc_size *out_size)
+{
+   if (!out_size)
```
Cloudef: assert instead, as out_size is not optional parameter at all to this. Though I'm questioning the usefulness of this function. It also might make sense to zero initialize out_size. It would make possible to use this function then regardless of return value as long as zero size is fine.

But for now, I think this function can be removed as I don't think of it as neccessity, writing wrapper function should be straight forward.

src/compositor/seat/pointer.c
```diff
@@ -139,10 +139,10 @@ pointer_paint(struct wlc_pointer *pointer, struct wlc_output *output)
...
    const struct wlc_point pos = {
-      chck_clamp(pointer->pos.x, 0, output->resolution.w),
-      chck_clamp(pointer->pos.y, 0, output->resolution.h)
+      chck_clamp(pointer->pos.x*scale, 0, output->resolution.w),
+      chck_clamp(pointer->pos.y*scale, 0, output->resolution.h)
```
Cloudef: Whitespace around operators.

src/compositor/seat/pointer.c
```diff
@@ -225,7 +227,7 @@ focus_view(struct wlc_pointer *pointer, struct wlc_surface *surf, wlc_handle old
          wlc_log(WLC_LOG_WARN, "Failed to push focused pointer resource to pool (out of memory?)");
 
       uint32_t serial = wl_display_next_serial(wlc_display());
-      wl_pointer_send_enter(wr, serial, surface, wl_fixed_from_double(pos->x), wl_fixed_from_double(pos->y));
+      wl_pointer_send_enter(wr, serial, surface, wl_fixed_from_double(pos->x/scale), wl_fixed_from_double(pos->y/scale));
```
Cloudef: I will skip the view.c and pointer.c code changes and detail my thoughts about how this functionality could be achieved with less changes in separate comment.

src/compositor/view.c
```diff
@@ -205,8 +207,8 @@ wlc_view_get_bounds(struct wlc_view *view, struct wlc_geometry *out_bounds, stru
       // xdg-surface client that draws drop shadows or other stuff.
       out_bounds->origin.x -= view->surface_commit.visible.origin.x;
       out_bounds->origin.y -= view->surface_commit.visible.origin.y;
-      out_bounds->size.w += surface->size.w - view->surface_commit.visible.size.w;
-      out_bounds->size.h += surface->size.h - view->surface_commit.visible.size.h;
+      out_bounds->size.w += surface->size.w / scale - view->surface_commit.visible.size.w;
+      out_bounds->size.h += surface->size.h / scale - view->surface_commit.visible.size.h;
```
Cloudef: Why is this necessary ?

src/compositor/view.c
```diff
@@ -215,9 +217,13 @@ wlc_view_get_bounds(struct wlc_view *view, struct wlc_geometry *out_bounds, stru
    if (!out_visible)
       return;
 
+   struct wlc_size temp=surface->size;
+   temp.w /= scale;
+   temp.h /= scale;
```
Cloudef: Ditto?

src/platform/render/gles2.c
```diff
@@ -734,6 +734,18 @@ view_paint(struct ctx *context, struct wlc_view *view)
...
+   wlc_handle output = wlc_view_get_output(convert_to_wlc_handle(view));
+   int32_t scale=wlc_output_get_scale(output);
+   geometry.origin.x*=scale;
+   geometry.origin.y*=scale;
+   geometry.size.w*=scale;
+   geometry.size.h*=scale;
+   settings.visible.origin.x*=scale;
+   settings.visible.origin.y*=scale;
+   settings.visible.size.w*=scale;
+   settings.visible.size.h*=scale;
```
Cloudef: Whitespaces around operations.

src/resources/types/surface.c
```diff
@@ -54,6 +54,9 @@ commit_state(struct wlc_surface *surface, struct wlc_surface_state *pending, str
    state_set_buffer(out, convert_from_wlc_resource(pending->buffer, "buffer"));
    state_set_buffer(pending, NULL);
 
+   out->scale = pending->scale;
```
Cloudef: out->scale = chck_max32(pending->scale, 1);

src/resources/types/surface.c
```diff
@@ -391,6 +394,7 @@ wlc_surface(struct wlc_surface *surface)
        !chck_iter_pool(&surface->subsurface_list, 4, 0, sizeof(wlc_resource)))
       goto fail;
 
+   surface->commit.scale = 1;
```
Cloudef: Is this neccessary?

src/compositor/seat/pointer.c
Cloudef: whietspace around operations.

src/compositor/seat/pointer.c
```diff
@@ -344,7 +350,7 @@ wlc_pointer_motion(struct wlc_pointer *pointer, uint32_t time, bool pass)
       if (!(wr = wl_resource_from_wlc_resource(*r, "pointer")))
          continue;
 
-      wl_pointer_send_motion(wr, time, wl_fixed_from_double(d.x), wl_fixed_from_double(d.y));
+      wl_pointer_send_motion(wr, time, wl_fixed_from_double(d.x/scale), wl_fixed_from_double(d.y/scale));
```
Cloudef: whitespace around oeprators.

src/compositor/seat/seat.c
```diff
@@ -145,14 +145,15 @@ input_event(struct wl_listener *listener, void *data)
...
+   const int32_t scale=output->information.scale;
    switch (ev->type) {
       case WLC_INPUT_EVENT_MOTION:
       {
          const struct wlc_size resolution = (output ? output->resolution : wlc_size_zero);
 
          const struct wlc_pointer_origin pos = {
-            chck_clamp(seat->pointer.pos.x + ev->motion.dx, 0, resolution.w),
-            chck_clamp(seat->pointer.pos.y + ev->motion.dy, 0, resolution.h),
+            chck_clamp(seat->pointer.pos.x + ev->motion.dx, 0, resolution.w/scale),
```
Cloudef: I would imagine the motion would need to be divided not the boundary.

src/compositor/seat/seat.c
```diff
@@ -165,8 +166,8 @@ input_event(struct wl_listener *listener, void *data)
          const struct wlc_size resolution = (output ? output->resolution : wlc_size_zero);
 
          const struct wlc_pointer_origin pos = {
-            ev->motion_abs.x(ev->motion_abs.internal, resolution.w),
-            ev->motion_abs.y(ev->motion_abs.internal, resolution.h)
+            ev->motion_abs.x(ev->motion_abs.internal, resolution.w/scale),
+            ev->motion_abs.y(ev->motion_abs.internal, resolution.h/scale)
```
Cloudef: Ditto in the cases here and below.

Cloudef:
Does the client expect coordinates it receives to be as if they were on 1x scale. That is we need to map for example 2x, mouse movement over surface to 1x scale coordinate system?       

Looking at this PR, it may be enough to simply just handle scaling in the wlc_view_get_bounds and wlc_view_get_visible_bounds function, and everything else will work as normal (hit testing, culling, rendering, input..) As input already does coordinate scaling to _surface size_ which is the 1x size without scaling. Can you try this and report that other changes are not needed?

For other things, I mentioned above wlc_output_set_scale sounds weird, but perhaps it can be useful to set custom scale on output. However does EDID / drm / etc.. give this information for us? Or does weston simply just use scale parameter in ini configuration as well.

src/resources/types/surface.c
ss1h2a3tw: Because commit scale is zero at the begining , it will cause dividing by zero.

ss1h2a3tw:
I will test it later, thanks for your reply.

src/resources/types/surface.c
Cloudef: Fair enough. It sounds like 0 scale for initial doesn't make much sense anyways. We consider and check client set surface scale at <= 0 illegal anyways.

ss1h2a3tw:
As only modifying get_bound result in the cursor's position differ between rendered on screen and inside client.
img_20160622_220356

in the photo , the cursor is in the title bar of weston-terminal (right middle in "progs"),
but the play button in the rhythmbox is highlighted.

using the code below
https://github.com/Cloudef/wlc/compare/master...ss1h2a3tw:test

src/compositor/view.c
ss1h2a3tw: surface's size seems to be the geometry*buffer_scale , but the bounds' coordinate should be  same as geometry?

src/compositor/view.c
ss1h2a3tw: I'm trying to just memcpy visible part when scaling the surface.

ss1h2a3tw:
@Cloudef 
Maybe changing line number 290 of surface.c can help?
It seems like it is directly copy the size of buffer, which explains why I need to divide surface size by scale.

Cloudef:
Dividing surface's size is enough for eveything to work? That sounds like good idea to me if so. Do we actually need the 2x size information anywhere now?

ss1h2a3tw:
Yeah, looks like buffer_scale can be handled as dividing surface's size.
supporting output scale will need to change seat/pointer.c , gles2.c

Cloudef:
Ok, so the steps we have done here.

1) Divide surface size with its scale (1 == unaffected, 2 == half, so on..)
2) Multiply input coordinates with output's scale, divide boundary by output scale. 
3) Render surfaces by multiplying surface geometry with output scale. (1 == unaffected, 2 == multiple of size, so on..)

1 and 3 seems like good idea as it avoids conditional calculation depending if surface size is less or more than output scale and so on.

What's the purpose of step 2? How does this work with scale 1 surfaces on scale 2 output? I assume the client treats coordinates in its native coordinate space, not in the scale 2 coordinate space. I assume this is just done so that scale 2 surfaces work on scale 2 output (?), if so maybe the better way is to modify the coordinate_transform structure in surface.c instead. 

Apart from the coordinate transformations still being unclear and the need of get_scaled_output_size() I think we are on right track.

ss1h2a3tw:
The reason of dividing boundary is that client use the coordinate of their surface,
when output scale=2, client is using the coordinate of resolution dividing the output scale.
So the pointer coordinate is half of output, when printing the cursor, we need to multiply by output scale.

And the input coordinate is right with my code.
img_20160623_175224
img_20160623_175333

weston-terminal scale=1
rhythmbox scale=2

Cloudef:
Could the output scale taken into account in surface's coordinate_transform when it's attached to output here: https://github.com/Cloudef/wlc/blob/35caed59fa992a82774711622d86ac1172abd07b/src/resources/types/surface.c#L299 ?

ss1h2a3tw:
OK I will try it later.

Cloudef:
Also, probably good idea to include scale = 3 surface in your test too, this should in theory be downscaled to scale 2 output.

ss1h2a3tw:
After my test,modify coordinate transform doesn't work.
Since
https://github.com/Cloudef/wlc/blob/309ddb03ee91b66ad4189d29ee651ac532d266a5/src/compositor/seat/pointer.c#L246
Subtracting offset first will make the offset been scaled by output scale too.

Since buffer_scale is informed by client,
I need to find some test client that I can define its buffer scale, or need some dirty hacking to test it.

Cloudef:
Probably best way is to modify e.g. weston-simple-shm https://cgit.freedesktop.org/wayland/weston/tree/clients/simple-shm.c#n316, you may send custom scale with wl_surface_set_buffer_scale. I'll test this PR too to get sense how the coordinate transformations go, and that the visual coordinates match the logical ones, (even if client gets correct coordinates, it's easy to break coordinates compositor side, so hit detection, culling etc fails)

ss1h2a3tw:
I found that simple-shm is using compositor version 1 so I need to change not only wl_surface_set_buffer_scale,
and I found simple-damage support scale as argument,
So I'm using it to test downscaling , seems working perfectly.

img_20160624_232212

Cloudef:
Ok that seems quite promising. I'll pull the code and test too.

ss1h2a3tw:
After my test of output scale = 3,4 upscale is working well as 2,
but because of the pointer coordinate is divide by output scale, it need move at least 3 pixels when output scale = 3, 4 pixels when output scale =4 and so on 

Maybe there's a better way for handling pointer coordinate.

Cloudef:
I locally managed to do this with only changing output.c and surface.c, I'm killing some other bugs at the same time, I'll post some diffs for testing then.

ss1h2a3tw:
I found that subsurface is rendering in the wrong position, I will fix it in few days.

Cloudef:
I already fixed that locally, and also the fact subsurfaces were not scaled properly when buffer size is being stretched (due to minimum size). I'll try to get my code out this week for testing.

ss1h2a3tw:
OK, Is there anything more I can help?

Cloudef:
I'll push the code for testing when ready. The work here is great resource on what needs to be done.

Cloudef:
https://github.com/Cloudef/wlc/tree/hidpi <- can you try this branch? For further fixes issues, probably good idea to open new PR against this.

ss1h2a3tw:
I have test it, and I found two problems.
When spawning a client which support buffer scale at scale = 1,then change output scale to 2, but it doesn't set the buffer scale into 2.

Another problem is not quite related to scale, when using weston-subsurface to test , the pointer event seems not sending or sending wrong when the cursor is on subsurface.

I will add some screenshot later.

Cloudef:
>When spawning a client which support buffer scale at scale = 1,then change output scale to 2, but it doesn't set the buffer scale into 2.

As far as wlc goes, it should send the new output information. It may be that the client just doesn't support this case.

>Another problem is not quite related to scale, when using weston-subsurface to test , the pointer event seems not sending or sending wrong when the cursor is on subsurface.

The pointer event problems may have been caused by unrelated change, that is basic support for "input area" of surfaces, but in sense it's incomplete as it only checks the area of views, not subsurfaces.

>The third problem is Xwayland seems not rendering in right scale.

I wonder if X11 has equivalent protocol to wayland scale that I should be aware of.. I really doubt it. Neverthless something to look into.

Drakulix:
X11 has no similar protocol I know of, but many programs respect the "Xft.dpi" variable you may set in .Xresources. Additionally many GUI frameworks have custom environment variables.

GDK/GTK: `GDK_SCALE=2`
Qt: `QT_AUTO_SCREEN_SCALE_FACTOR=2`
Elementary: `ELM_SCALE=2`

I do not know, if wlc is supposed to set environment variables, but there should be at least a callback for changed scale factors, so an actual compositor might set those variables.

Cloudef:
output.resolution callback already gets called, however it may be good idea to break the API of this callback and include from_scale and to_scale in the arguments as well. So you don't lose the information of old scale.

include/wlc/wlc.h
SirCmpwn: :+1: for this function

SirCmpwn:
Did some initial work on getting this branch working with sway. My branch is [here](https://github.com/SirCmpwn/sway/tree/hidpi). A few things I've noticed (some of this may be sway's fault and some of it may be wlc's fault, not sure yet):

* subsurfaces are 4x scaled if I use 2x scaling
* window borders are fucked up
* x windows are stretched rather than setting the xwayland DPI (this is a difficult one, since x only supports one DPI for the whole thing)

SirCmpwn:
[Screenshot](https://sr.ht/Eh8T.png)

Use `output [name] scale 2` or similar to set the scale factor on sway.

ss1h2a3tw:
@SirCmpwn  please use the branch https://github.com/Cloudef/wlc/tree/hidpi
my branch here have bug handling subsurface scale

SirCmpwn:
Tried again with wlc's hidpi branch, works better. Pushed a new version to my hidpi branch.

Cloudef:
Merged into master.

PR #176, opened 2016-07-11
Use a fallback XDG_RUNTIME_DIR, when applicable

Hummer12007:
This makes it easier to run the compositors for people not using ck/logind or anything else setting XDG_RUNTIME_DIR

### DIFF
--- a/src/wlc.c
+++ b/src/wlc.c
@@ -293,6 +293,18 @@ wlc_init(void)
       wlc.interface = old_interface;
    }
 
+   // set a fallback XDG_RUNTIME_DIR, if required (needed for libwayland)
+   if (!getenv("XDG_RUNTIME_DIR")) {
+      char buf[] = "/tmp/wlc.XXXXXX";
+      char *dir = mkdtemp(buf);
+      if (dir) {
+         wlc_log(WLC_LOG_WARN, "No XDG_RUNTIME_DIR set, creating a temporary directory");
+         setenv("XDG_RUNTIME_DIR", dir, false);
+      } else {
+         wlc_log(WLC_LOG_ERROR, "No XDG_RUNTIME_DIR set, could not set up a temporary directory");
+      }
+   }
+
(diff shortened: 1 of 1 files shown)


Earnestly:
This needs a little more work than simply using `/tmp`.

As a good citizen it should ideally check the `TMPDIR` environment first and use that before falling back on `/tmp`.

To be more closely related to what `RUNTIME_DIR` is doing (and it's fundamental purpose (which is basically emulating namespaces badly)) one should at least create a subdirectory under `/tmp` with the user as the owner with a `0700` permission mask.  There are other requirements the XDG spec imposes on `RUNTIME_DIR` but most of them are pretty irrelevant.

src/wlc.c
```diff
@@ -293,6 +293,18 @@ wlc_init(void)
       wlc.interface = old_interface;
    }
 
+   // set a fallback XDG_RUNTIME_DIR, if required (needed for libwayland)
+   if (!getenv("XDG_RUNTIME_DIR")) {
+      char buf[] = "/tmp/wlc.XXXXXX";
+      char *dir = mkdtemp(buf);
+      if (dir) {
+      	wlc_log(WLC_LOG_WARN, "No XDG_RUNTIME_DIR set, creating a temporary directory");
+      	setenv("XDG_RUNTIME_DIR", dir, false);
```
Cloudef: Fix the indents (wrong level).

Hummer12007:
done

Cloudef:
Thanks

pjanouch:
This is just purely dysfunctional, since it does the following:
 - create a directory as root (otherwise we can't `VT_ACTIVATE`)
 - drop privileges
 - `wl_display_add_socket_auto()` can't write into it
 - `Failed to add socket to wayland display`

because you forgot to address this:

> create a subdirectory under /tmp with the user as the owner and a 0700 permission mask

Cloudef:
Time to revert it then.

PR #177, opened 2016-07-29
implement wlc_view_get_pid()

zandrmartin:
Hello! Please review my attempt at implementing a `wlc_view_get_pid()` function. We need this in Sway (specifically for [this issue](https://github.com/SirCmpwn/sway/issues/462)). I'm very open to suggestions/critiques on this - I am pretty new to C and WLC is a bit more complicated than anything I've dealt with before. Thanks!

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -8,6 +8,7 @@ extern "C" {
 #include <wlc/defines.h>
 #include <wlc/geometry.h>
 #include <xkbcommon/xkbcommon-keysyms.h>
+#include <sys/types.h>
 
 struct wlc_event_source;
 
@@ -61,6 +62,7 @@ enum wlc_view_property_update_bit {
    WLC_BIT_PROPERTY_TITLE = 1<<0,
    WLC_BIT_PROPERTY_CLASS = 1<<1,
    WLC_BIT_PROPERTY_APP_ID = 1<<2,
+   WLC_BIT_PROPERTY_PID = 1<<3,
 };
 
 /** wlc_view_set_geometry(); Edges in interface interface.view.request.resize function. */
@@ -387,6 +389,9 @@ const char* wlc_view_get_class(wlc_handle view);
 
 /** Get app id. (xdg-surface only) */
 const char* wlc_view_get_app_id(wlc_handle view);
+
+/** Get pid. */
+pid_t wlc_view_get_pid(wlc_handle view);
 
 /** --  Input API
  * Very recent stuff, things may change.
--- a/src/compositor/view.c
+++ b/src/compositor/view.c
@@ -440,6 +440,16 @@ wlc_view_set_app_id_ptr(struct wlc_view *view, const char *app_id)
 }
 
 void
+wlc_view_set_pid_ptr(struct wlc_view *view, pid_t pid)
+{
+   if (!view || view->data.pid == pid)
+      return;
+
+   view->data.pid = pid;
+   WLC_INTERFACE_EMIT(view.properties_updated, convert_to_wlc_handle(view), WLC_BIT_PROPERTY_PID);
+}
+
+void
 wlc_view_close_ptr(struct wlc_view *view)
 {
    if (!view)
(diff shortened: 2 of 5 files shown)


SirCmpwn:
LGTM :+1:

include/wlc/wlc.h
```diff
@@ -388,6 +390,8 @@ const char* wlc_view_get_class(wlc_handle view);
 /** Get app id. (xdg-surface only) */
 const char* wlc_view_get_app_id(wlc_handle view);
 
+/** Get pid. */
+pid_t wlc_view_get_pid(wlc_handle view);
```
Cloudef: Separate this line from the comments below.

src/compositor/view.c
```diff
@@ -440,6 +441,15 @@ wlc_view_set_app_id_ptr(struct wlc_view *view, const char *app_id)
 }
 
 void
+wlc_view_set_pid(struct wlc_view *view, pid_t pid)
```
Cloudef: wlc_view_set_pid_ptr as this works on the pointer not handle.

src/compositor/view.c
```diff
@@ -440,6 +441,15 @@ wlc_view_set_app_id_ptr(struct wlc_view *view, const char *app_id)
...
+{
+   if (view) {
```
Cloudef: Use precondition instead, that is if (!view) return; to avoid nesting ifs.

src/compositor/view.c
```diff
@@ -649,6 +659,19 @@ wlc_view_get_app_id(wlc_handle view)
    return get_cstr(convert_from_wlc_handle(view, "view"), offsetof(struct wlc_view, data.app_id));
 }
 
+WLC_API pid_t
+wlc_view_get_pid(wlc_handle handle)
+{
+   struct wlc_view *view = convert_from_wlc_handle(handle, "view");
```
Cloudef: view should be checked for NULL.

src/xwayland/xwm.c
```diff
@@ -221,6 +221,7 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
          wlc_dlog(WLC_DBG_XWM, "WM_TRANSIENT_FOR: %u", *xid);
       } else if (props[i] == x11.atoms[NET_WM_PID] && reply->type == XCB_ATOM_CARDINAL) {
          // PID
+         wlc_view_set_pid(view, *(pid_t *)xcb_get_property_value(reply));
```
Cloudef: Seems to identical to weston. Weston seems to do some validation on the pid though, not sure if that's neccessary. Not sure about the inner workings of xcb_get_property_value, but I hope it returns data with sizeof(1) since otherwise this could cause aliasing issues and thus memcpy would be correct way to get the data. But since weston does it the same way I assume it's ok.

src/compositor/view.c
```diff
@@ -2,6 +2,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <assert.h>
+#include <sys/types.h>
```
Cloudef: This already comes from view.h that is included above. So probably not needed. It's needed in header since pid_t is exposed.

src/xwayland/xwm.c
upwhere: Jari Vetoniemi writes:
> > @@ -221,6 +221,7 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
> >           wlc_dlog(WLC_DBG_XWM, "WM_TRANSIENT_FOR: %u", *xid);
> >        } else if (props[i] == x11.atoms[NET_WM_PID] && reply->type == XCB_ATOM_CARDINAL) {
> >           // PID
> > +         wlc_view_set_pid(view, *(pid_t *)xcb_get_property_value(reply));
> 
> [...] but I hope it returns data with sizeof(1) since
> otherwise this could cause aliasing issues and thus memcpy
> would be correct way to get the data. But since weston does
> it the same way I assume it's ok.

Isn't meta-information like this great to interleave as a
comment? The next person editing this code or trying to fix a
bug might trip on it.

PR #182, opened 2016-08-13
Changes to enable building on FreeBSD. 

yohanesu75:
Need to confirm still builds and works as expected on Linux!

### DIFF
--- a/src/session/fd.c
+++ b/src/session/fd.c
@@ -82,6 +82,7 @@ write_fd(int sock, int fd, const void *buffer, ssize_t buffer_size)
 {
    char control[CMSG_SPACE(sizeof(int))];
    memset(control, 0, sizeof(control));
+
    struct msghdr message = {
       .msg_name = NULL,
       .msg_namelen = 0,
@@ -90,18 +91,19 @@ write_fd(int sock, int fd, const void *buffer, ssize_t buffer_size)
          .iov_len = buffer_size
       },
       .msg_iovlen = 1,
-      .msg_control = control,
-      .msg_controllen = sizeof(control),
+      .msg_control = NULL,
+      .msg_controllen = 0,
    };
 
-   struct cmsghdr *cmsg = CMSG_FIRSTHDR(&message);
    if (fd >= 0) {
+      message.msg_control = control;
+      message.msg_controllen = sizeof(control);
+
+      struct cmsghdr *cmsg = CMSG_FIRSTHDR(&message);
       cmsg->cmsg_len = CMSG_LEN(sizeof(fd));
       cmsg->cmsg_level = SOL_SOCKET;
       cmsg->cmsg_type = SCM_RIGHTS;
       memcpy(CMSG_DATA(cmsg), &fd, sizeof(fd));
-   } else {
-      cmsg->cmsg_len = CMSG_LEN(0);
    }
 
    return sendmsg(sock, &message, 0);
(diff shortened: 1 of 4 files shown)


src/session/fd.c
```diff
@@ -132,6 +134,9 @@ recv_fd(int sock, int *out_fd, void *out_buffer, ssize_t buffer_size)
    if ((read = recvmsg(sock, &message, 0)) < 0)
       return read;
 
+   if(message.msg_controllen == 0)
```
Cloudef: Add space between if and (

src/session/tty.c
```diff
@@ -147,6 +165,79 @@ setup_tty(int fd, bool replace_vt)
...
+      die("Could not get current vt");
+
+   wlc.old_state.vt = state.v_active;
+
+   if (ioctl(fd, VT_ACTIVATE, wlc.vt) == -1)
+      die("Could not activate vt%d", wlc.vt);
+
+   if (ioctl(fd, VT_WAITACTIVE, wlc.vt) == -1)
+      die("Could not wait for vt%d to become active", wlc.vt);
+#endif
```
Cloudef: Maybe just remove this code, and add whenever it has something similar.

src/xwayland/xwayland.c
```diff
@@ -287,10 +287,15 @@ wlc_xwayland_init(void)
...
 
+#ifdef HAVE_CLEARENV
       if (clearenv() != 0) {
          wlc_log(WLC_LOG_WARN, "Failed to clear environment");
          _exit(EXIT_FAILURE);
       }
+#else
+      extern char **environ;
+      environ[0] = NULL;
+#endif
```
Cloudef: Can you turn these instances into own function, so there is no need for conditional #if's each time.

yohanesu75: Would defining a macro in internal.h be OK for this?

src/macros.h
```diff
@@ -22,4 +22,10 @@
 
 #define except(x) if (!(x)) { wlc_log(WLC_LOG_ERROR, "assertion failed: %s", #x); abort(); }
 
+#ifdef HAVE_CLEARENV
+#define CLEARENV() if (clearenv() != 0) { die("Failed to clear environment"); }
+#else
+#define CLEARENV() extern char **environ; environ[0] = NULL;
+#endif
```
Cloudef: I would prefer static inline functions instead of macros, maybe in internal.h as you suggested.

src/xwayland/xwayland.c
```diff
@@ -287,15 +287,7 @@ wlc_xwayland_init(void)
...
-#ifdef HAVE_CLEARENV
-      if (clearenv() != 0) {
-         wlc_log(WLC_LOG_WARN, "Failed to clear environment");
-         _exit(EXIT_FAILURE);
-      }
-#else
-      extern char **environ;
-      environ[0] = NULL;
-#endif
+	  CLEARENV();
```
Cloudef: The indent here and place before is off.

yohanesu75:
There, how is that?

Cloudef:
Looks good, squash and I'll merge.

src/session/fd.c
```diff
@@ -90,18 +91,19 @@ write_fd(int sock, int fd, const void *buffer, ssize_t buffer_size)
...
+      .msg_control = NULL,
+      .msg_controllen = 0,
    };
 
-   struct cmsghdr *cmsg = CMSG_FIRSTHDR(&message);
    if (fd >= 0) {
+      message.msg_control = control;
+      message.msg_controllen = sizeof(control);
+
+	  struct cmsghdr *cmsg = CMSG_FIRSTHDR(&message);
```
SirCmpwn: Indent issues

Cloudef: Thanks, I missed this.

yohanesu75: Sorry will fix. My .dir-locals.el wasn't working properly.

PR #187, opened 2016-08-20
Improved fd and tty code for FreeBSD.

yohanesu75:
Also did  a re-indent on the whole file so a bunch of tabs where replaces with spaces. 

NOTE: Not tested on Linux!


yohanesu75:
Need some more work. Closing for now.

PR #192, opened 2016-08-29
Fix: Inform X11 windows about border size. Fixes #191

### LINKED ISSUE #191: Accepting requested geometry causes X windows to grow
Hello,

I found strange issue with wlc_set_view_request_geometry_cb callback.

As per documentation, I'm trying to agree on requested geometry by setting it using wlc_view_set_geometry. But doing so apparently sets size of X window larger by 2px in both directions, which in turn causes new size to be requested, ending in infinite loop of window slowly growing.

[Here is minimal example](http://pastebin.com/7gYUKA4x). Compiling it with ``` gcc test.c -lwlc -o test ```, running generated binary and then launching `DISPLAY=:1 xterm` from another console replicates this problem.

kozec:
Some, or maybe even all X11 windows are having trouble when wlc_set_view_request_geometry_cb callback accepts geometry they requested, as this geometry is then send back with incorrect border size set. This cases affected X11 windows to slowly, uncontrollably and indefinitely grow.

### DIFF
--- a/src/xwayland/xwm.c
+++ b/src/xwayland/xwm.c
@@ -288,8 +288,8 @@ static void
 set_geometry(xcb_window_t window, const struct wlc_geometry *g)
 {
    assert(g);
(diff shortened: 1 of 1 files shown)


Cloudef:
Thanks

PR #194, opened 2016-09-03
Improvements to FreeBSD tty code.

johalun:


Cloudef:
There seems to be quite lots of indent / space changes, can you fix those before I check further.

yohanesu75:
Ah you're right.. I thought it was tabs that had been converted to spaces automatically but I see that indent is changed. Will fix it!

yohanesu75:
We end up with#ifdefs all over the place.. Maybe cleanest to put in separate files?

Cloudef:
You could separate the parts that are different into functions into own files, provide tty_backend struct that contains the function prototypes. Have something like wlc_tty_backend_init(&backend); at begin of wlc_tty_init, the implementation that gets called for wlc_tty_backend_init would depend on platform (linux or bsd), by telling cmake which .c file to compile. tty.h would contain the prototype for wlc_tty_backend_init, but implementation would be in the platform specific file.

If that makes sense. Above is also really easy to modify so that tty backends could be loaded from .so's (plugins), but I don't think we ever need that. (Or at least I hope there doesn't exist multiple tty apis on single platform)

SirCmpwn:
Presumably this is obsolete.

Issue #174, opened 2016-06-23
Support for color temperature


psgarsenal:
Hi, one of the only things lacking for me is the support for things like f.lux or redshift.
As said here: jonls/redshift#55 the support has to be implemented at your level (as I understand it, I might be wrong).
Do you think this is possible?
Thanks for reading!

Cloudef:
It was something I planned way back. There is @giucam who did protocol and modified redshift to use it. So as long as this protocol <https://github.com/giucam/orbital/blob/gamma/protocol/gamma-control.xml> is implemented in compositor it would work.

giucam:
Note however that since the protocol is currently only in orbital i didn't try to upstream the redshift changes. So they are available on https://github.com/giucam/redshift

psgarsenal:
Ok then going to check on sway side to see if ti is possible, thanks for the answers!

Cloudef:
@psgarsenal It might make more sense to have this protocol inside wlc though, so sway gets it for free.

psgarsenal:
I guess we could ask @SirCmpwn on his thoughts about this

SirCmpwn:
Should go in wlc.

embik:
This issue should be re-opened, right? It seems to be the "correct" issue for color temperature / redshift functionality. I'd really like to see this in wlc (and therefore in sway).

craftyguy:
Any update on this?

PR #216, opened 2016-12-28
Add gamma control API

SirCmpwn:
Closes https://github.com/Cloudef/wlc/issues/174

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -280,6 +280,12 @@ bool wlc_output_get_sleep(wlc_handle output);
 /** Wake up / sleep. */
 void wlc_output_set_sleep(wlc_handle output, bool sleep);
 
+/** Set gamma. R, G, and B are color ramp arrays of size elements. */
+void wlc_output_set_gamma(wlc_handle output, uint16_t size, uint16_t *r, uint16_t *g, uint16_t *b);
+
+/** Get gamma size */
+uint16_t wlc_output_get_gamma_size(wlc_handle output);
+
 /**
  * Get real resolution.
  * Resolution applied by either wlc_output_set_resolution call or initially.
--- a/src/compositor/output.c
+++ b/src/compositor/output.c
@@ -818,6 +818,16 @@ wlc_output_set_sleep_ptr(struct wlc_output *output, bool sleep)
 }
 
 void
+wlc_output_set_gamma_ptr(struct wlc_output *output, uint16_t size, uint16_t *r, uint16_t *g, uint16_t *b)
+{
+   if (!output)
+      return;
+
+   if (output->bsurface.api.set_gamma)
+      output->bsurface.api.set_gamma(&output->bsurface, size, r, g, b);
+}
+
+void
 wlc_output_set_mask_ptr(struct wlc_output *output, uint32_t mask)
 {
    if (!output)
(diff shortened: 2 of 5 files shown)


src/compositor/output.c
```diff
@@ -913,6 +923,23 @@ wlc_output_set_sleep(wlc_handle output, bool sleep)
...
+wlc_output_set_gamma(wlc_handle output, uint16_t size, uint16_t *r, uint16_t *g, uint16_t *b)
+{
+   wlc_output_set_gamma_ptr(convert_from_wlc_handle(output, "output"), size, r, g, b);
+}
+
+WLC_API uint32_t
+wlc_output_get_gamma_size(wlc_handle output)
+{
+   struct wlc_output *_output = convert_from_wlc_handle(output, "output");
+
```
Cloudef: precondition here if (!_output || !_output->bsurface.api.get_gamma_size) return 0;

include/wlc/wlc.h
```diff
@@ -280,6 +280,12 @@ bool wlc_output_get_sleep(wlc_handle output);
 /** Wake up / sleep. */
 void wlc_output_set_sleep(wlc_handle output, bool sleep);
 
+/** Set gamma */
+void wlc_output_set_gamma(wlc_handle output, uint16_t size, uint16_t *r, uint16_t *g, uint16_t *b);
```
Cloudef: Just wondering why the rgb are pointers? Arrays?

src/compositor/output.c
```diff
@@ -913,6 +923,23 @@ wlc_output_set_sleep(wlc_handle output, bool sleep)
...
+   if (_output->bsurface.api.get_gamma_size)
+      return _output->bsurface.api.get_gamma_size(&_output->bsurface);
```
Cloudef: You can then take out this branching and return this directly.

include/wlc/wlc.h
Cloudef: Maybe document the units or something, or just refer to drm documentation since this API models it I guess.

include/wlc/wlc.h
```diff
@@ -280,6 +280,12 @@ bool wlc_output_get_sleep(wlc_handle output);
...
+
+/** Get gamma size */
+uint32_t wlc_output_get_gamma_size(wlc_handle output);
```
Cloudef: Is this same as size given in setter above? Why is it uint32_t here and uint16_t there? Same thing for the documentation here.

src/compositor/output.c
SirCmpwn: Won't that segfault if get_gamma_size is NULL for this particular backend (which in practice will happen for the x backend)

include/wlc/wlc.h
SirCmpwn: Yeah I dunno how gamma works, I just gave an interface on top of drm. Will mention that in the docs.

include/wlc/wlc.h
SirCmpwn: Good catch.

src/compositor/output.c
Cloudef: The precondition mentioned above will check both if output is null or if that function pointer is null.

SirCmpwn: Ah, got it.

src/compositor/output.c
```diff
@@ -929,15 +929,14 @@ wlc_output_set_gamma(wlc_handle output, uint16_t size, uint16_t *r, uint16_t *g,
...
+WLC_API uint16_t
 wlc_output_get_gamma_size(wlc_handle output)
 {
    struct wlc_output *_output = convert_from_wlc_handle(output, "output");
 
-   if (_output->bsurface.api.get_gamma_size)
-      return _output->bsurface.api.get_gamma_size(&_output->bsurface);
-   else
-      return 0;
+   if (!_output || _output->bsurface.api.get_gamma_size) return 0;
```
Cloudef: Add newline, I just inlined the code in comment before. Sorry about that.

Cloudef: You can squash the commits then, it looks good otherwise.

PR #222, opened 2017-01-07
Use WM_CLASS class instead of instance (fix #218)

### LINKED ISSUE #218: return second string of WM_CLASS
The function `wlc_view_get_class()` returns the first part of the `WM_CLASS` property while it's more likely that the second one is expected. Eg. i3wm uses the second string of `WM_CLASS` when using `for_window [class="my_application"]` for matching:

> `class`
> The X11 window class (second part of WM_CLASS)

(http://i3wm.org/docs/userguide.html#command_criteria)

This is the property as specified by the X.org foundation:

> **WM_CLASS Property**
> 
> The WM_CLASS property (of type STRING without control characters) contains two consecutive null-terminated strings. These specify the Instance and Class names to be used by both the client and the window manager for looking up resources for the application or as identifying information. This property must be present when the window leaves the Withdrawn state and may be changed only while the window is in the Withdrawn state. Window managers may examine the property only when they start up and when the window leaves the Withdrawn state, but there should be no need for a client to change its state dynamically.
> 
> The two strings, respectively, are:
> 
>  * A string that names the particular instance of the application to which the client that owns this window belongs. Resources that are specified by instance name override any resources that are specified by class name. Instance names can be specified by the user in an operating-system specific manner. On POSIX-conformant systems, the following conventions are used:
> 
>    * If "-name NAME" is given on the command line, NAME is used as the instance name.
> 
>    * Otherwise, if the environment variable RESOURCE_NAME is set, its value will be used as the instance name.
> 
>    * Otherwise, the trailing part of the name used to invoke the program (argv[0] stripped of any directory names) is used as the instance name. 
> 
>  * A string that names the general class of applications to which the client that owns this window belongs. Resources that are specified by class apply to all applications that have the same class name. Class names are specified by the application writer. Examples of commonly used class names include: "Emacs", "XTerm", "XClock", "XLoad", and so on. 
> 
> Note that WM_CLASS strings are null-terminated and, thus, differ from the general conventions that STRING properties are null-separated. This inconsistency is necessary for backwards compatibility. 
> 

https://www.x.org/releases/X11R7.7/doc/xorg-docs/icccm/icccm.html#WM_CLASS_Property

See also:
https://github.com/SirCmpwn/sway/issues/1020

LeonardKoenig:
Usually the second part of WM_CLASS (the actual window class, not the instance) is
expected (compare to i3wm class property) so we use this one.

This fixes #218

### DIFF
--- a/src/xwayland/xwm.c
+++ b/src/xwayland/xwm.c
@@ -205,7 +205,15 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
          // Class && Name
          // STRING == latin1, but we naively just read it as is. For full support we should convert to utf8.
          if (props[i] == XCB_ATOM_WM_CLASS) {
-            wlc_view_set_class_ptr(view, xcb_get_property_value(reply), xcb_get_property_value_length(reply));
+            char *class = xcb_get_property_value(reply);
+            size_t class_total_len = xcb_get_property_value_length(reply);
+
+            size_t class_class_offset = strnlen(class, class_total_len)+1;
+            if (class_class_offset >= class_total_len) {
+               wlc_view_set_class_ptr(view, class, class_total_len);
+	    } else {
+               wlc_view_set_class_ptr(view, class+class_class_offset, class_total_len-class_class_offset);
+	    }
             wlc_dlog(WLC_DBG_XWM, "WM_CLASS: %s", view->data._class.data);
          } else if (props[i] == XCB_ATOM_WM_NAME || props[i] == x11.atoms[NET_WM_NAME]) {
             if (reply->type != XCB_ATOM_STRING  || !win->has_utf8_title) {


src/xwayland/xwm.c
```diff
@@ -205,7 +205,11 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
          // Class && Name
          // STRING == latin1, but we naively just read it as is. For full support we should convert to utf8.
          if (props[i] == XCB_ATOM_WM_CLASS) {
-            wlc_view_set_class_ptr(view, xcb_get_property_value(reply), xcb_get_property_value_length(reply));
+            char *class = xcb_get_property_value(reply);
+            size_t class_total_len = xcb_get_property_value_length(reply);
+
+            size_t class_class_offset = strnlen(class, class_total_len)+1;
+            wlc_view_set_class_ptr(view, class+class_class_offset, class_total_len-class_class_offset);
```
Cloudef: The calculation for length here may result in -1 if the property doesn't have the second value.

LeonardKoenig: Does that matter, I think `wlc_view_set_class_ptr` would then just set it to `NULL`, doesn't it?

But yes, we possibly should set it to the first value instead as fallback.

LeonardKoenig: Currently not at home but will draft a patch soon.

Cloudef: >Does that matter, I think wlc_view_set_class_ptr would then just set it to NULL, doesn't it?

Can we guarantee that the value is always 0 terminated? If that assumption is not true then the class + class_class_offset may not result in NULL.

>But yes, we possibly should set it to the first value instead as fallback.

That's probably good idea.

LeonardKoenig: I think the checks implemented in the function would just not let it pass with `lenght < 0` so I guess it wouldn't matter.

Cloudef: The length is size_t so it won't hold negative numbers.

LeonardKoenig: you're right, it would just overflow.

LeonardKoenig:
Finally got around fixing this, what do you think?

src/xwayland/xwm.c
```diff
@@ -209,7 +209,11 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
             size_t class_total_len = xcb_get_property_value_length(reply);
 
             size_t class_class_offset = strnlen(class, class_total_len)+1;
-            wlc_view_set_class_ptr(view, class+class_class_offset, class_total_len-class_class_offset);
+            if (class_class_offset >= class_total_len) {
```
Cloudef: Won't this fail if class_total_len == 0?

src/xwayland/xwm.c
```diff
@@ -209,7 +209,11 @@ read_properties(struct wlc_xwm *xwm, struct wlc_x11_window *win, const xcb_atom_
...
+               wlc_view_set_class_ptr(view, class, class_class_offset-1);
```
Cloudef: Also just use the original len here.

src/xwayland/xwm.c
LeonardKoenig: Then `class_class_offset=1` and so `class_class_offset=0` which is what we intend.

src/xwayland/xwm.c
LeonardKoenig: This'd make for either an additional variable or an additional `strlen()` call, it doesn't really matter though

src/xwayland/xwm.c
Cloudef: Ah yeah, I see it now.

src/xwayland/xwm.c
Cloudef: I mean, you can just use the class_total_len as the second parameter.

LeonardKoenig: you're totally right. I just push the change.
Do you prefer me squashing the commits before merging?

Cloudef: Yup squash the commits.

LeonardKoenig:
Done squashing

SirCmpwn:
Would be nice to also be able to get the instance out of wlc.

LeonardKoenig:
This'd be the old functionality of the function, we'd just to add a new one, not difficult. The only question is whether to add it in here or regard it as a new feature and seperate it.

Cloudef:
I'll merge this now. Just do new PR if you decide to add instance.

PR #224, opened 2017-01-10
Include sys/sysmacros.h for major/minor(3) (Linux)

Hummer12007:
sys/sysmacros.h implicit inclusion is being deprecated (even with _DEFAULT_SOURCE), so it would cause build breakage on future updates.

See https://bugs.gentoo.org/show_bug.cgi?id=604608 and https://sourceware.org/ml/libc-alpha/2015-11/msg00253.html (also mentioned in glibc news)

### DIFF
--- a/src/session/fd.c
+++ b/src/session/fd.c
@@ -12,6 +12,7 @@
 #include <sys/types.h>
 #ifdef __linux__
 #include <linux/major.h>
+#include <sys/sysmacros.h>
 #endif
 #include <xf86drm.h>
 #include "internal.h"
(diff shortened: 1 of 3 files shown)


Cloudef:
Do you know if this breaks compile anywhere?

Hummer12007:
Well, it seems to break the build in gentoo with glibc-2.24 (which is not even keyworded yet), as they decided to patch the implicit include out (again). Same thing happened with 2.23 before (but they have decided to revert it before stabilizing). Who knows what'll happen this time...

Cloudef:
Lets merge this and think about it if there are complaints.

PR #233, opened 2017-01-31
wlc: Use eglGetPlatformDisplayEXT where available

jwrdegoede:
eglGetDisplay does not know what sort of handle it is getting passed
in (a gbm device handle or an X11 display pointer) on systems without
glvnd some poking into internal X11 / gbm structures is used to figure
this out, which is somewhat dangerous but works.

On system using glvnd this does not work and eglGetDisplay returns NULL.

This commit makes wlc track the EGLNativeDisplayType pointer's
underlying type and uses eglGetPlatformDisplayEXT where available so
that the EGL implementation does not need to guess the type of the
handle it is being passed.

BugLink: https://bugzilla.redhat.com/show_bug.cgi?id=1413579
Signed-off-by: Hans de Goede <hdegoede@redhat.com>

### DIFF
--- a/src/platform/context/egl.c
+++ b/src/platform/context/egl.c
@@ -148,6 +148,47 @@ terminate(struct ctx *context)
    free(context);
 }
 
+/*
+ * Taken from glamor_egl.c from the Xorg xserver, which is MIT licensed like wlc
+ *
+ * Create an EGLDisplay from a native display type. This is a little quirky
+ * for a few reasons.
+ *
+ * 1: GetPlatformDisplayEXT and GetPlatformDisplay are the API you want to
+ * use, note these have different function signatures in the third argument.
+ *
+ * 2: You can't tell whether you have EGL 1.5 at this point, because
+ * eglQueryString(EGL_VERSION) is a property of the display, which we don't
+ * have yet. So you have to query for extensions no matter what.
+ *
+ * 3. There is no EGL_KHR_platform_base to complement the EXT one, thus one
+ * needs to know EGL 1.5 is supported in order to use the eglGetPlatformDisplay
+ * function pointer.
+ *
+ * We can workaround this (circular dependency) by probing for the EGL 1.5
+ * platform extensions (EGL_KHR_platform_gbm and friends) yet it doesn't seem
+ * like mesa will be able to adverise these (even though it can do EGL 1.5).
+ */
+static EGLDisplay
+get_display(struct ctx *context, EGLint type, void *native)
+{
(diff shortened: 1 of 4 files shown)


src/platform/context/egl.c
```diff
@@ -148,6 +148,47 @@ terminate(struct ctx *context)
...
+ * function pointer.
+ *
+ * We can workaround this (circular dependency) by probing for the EGL 1.5
+ * platform extensions (EGL_KHR_platform_gbm and friends) yet it doesn't seem
+ * like mesa will be able to adverise these (even though it can do EGL 1.5).
+ */
+static EGLDisplay
+get_display(struct ctx *context, EGLint type, void *native)
+{
+   PFNEGLGETPLATFORMDISPLAYEXTPROC getPlatformDisplayEXT;
```
Cloudef: Was this accidental duplicate definition? Otherwise this looks quite good. Scary but good :)

jwrdegoede:
Hi,

On 31-01-17 15:07, Jari Vetoniemi wrote:
> *@Cloudef* commented on this pull request.
>
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>
> In src/platform/context/egl.c <https://github.com/Cloudef/wlc/pull/233#pullrequestreview-19320318>:
>
>> + * 2: You can't tell whether you have EGL 1.5 at this point, because
> + * eglQueryString(EGL_VERSION) is a property of the display, which we don't
> + * have yet. So you have to query for extensions no matter what.
> + *
> + * 3. There is no EGL_KHR_platform_base to complement the EXT one, thus one
> + * needs to know EGL 1.5 is supported in order to use the eglGetPlatformDisplay
> + * function pointer.
> + *
> + * We can workaround this (circular dependency) by probing for the EGL 1.5
> + * platform extensions (EGL_KHR_platform_gbm and friends) yet it doesn't seem
> + * like mesa will be able to adverise these (even though it can do EGL 1.5).
> + */
> +static EGLDisplay
> +get_display(struct ctx *context, EGLint type, void *native)
> +{
> +   PFNEGLGETPLATFORMDISPLAYEXTPROC getPlatformDisplayEXT;
>
> Was this accidental duplicate definition? Otherwise this looks quite good. Scary but good :)

Yes, that was a copy paste mistake the top level one can be removed.

Regards,

Hans

PR #245, opened 2017-03-30
Add EGL stream support

vially:
Add support for EGL streams used by the proprietary NVIDIA driver: #138

The current implementation selects the EGL stream buffer API (instead of the default GBM one) when the `WLC_USE_EGLDEVICE` environment variable is not empty (e.g.: `export WLC_USE_EGLDEVICE=1` is needed at runtime).

I've only tested it by running the `example` compositor on a single monitor setup (Archlinux using latest NVIDIA drivers) and it seems to work.

Disclaimer: I don't know `C`, so most probably the code contains lots of bugs so please review very carefully (especially around memory allocations/deallocations).

### DIFF
--- a/src/platform/backend/drm.c
+++ b/src/platform/backend/drm.c
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <sys/mman.h>
 #include <sys/select.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
@@ -17,6 +18,7 @@
 #include "backend.h"
 #include "compositor/compositor.h"
 #include "compositor/output.h"
+#include "platform/context/egl.h"
 #include "session/fd.h"
 
 // FIXME: Contains global state (event_source && fd)
@@ -27,13 +29,14 @@ struct drm_output_information {
    drmModeConnector *connector;
    drmModeEncoder *encoder;
    drmModeCrtc *crtc;
+   drmModeModeInfo mode;
    struct wlc_output_information info;
    uint32_t width, height;
 };
 
 struct drm_surface {
-   struct gbm_device *device;
-   struct gbm_surface *surface;
+   void *device;
+   struct gbm_surface *gbm_surface;
    drmModeConnector *connector;
    drmModeEncoder *encoder;
    drmModeCrtc *crtc;
@@ -50,10 +53,8 @@ struct drm_surface {
 };
 
 static struct {
-   struct gbm_device *device;
-} gbm;
-
-static struct {
+   bool use_egldevice;
+   void *device;
    int fd;
    struct wl_event_source *event_source;
 } drm;
@@ -61,12 +62,12 @@ static struct {
 static void
 release_fb(struct gbm_surface *surface, struct drm_fb *fb)
 {
-   assert(surface && fb);
+   assert(fb);
 
    if (fb->fd > 0)
       drmModeRmFB(drm.fd, fb->fd);
 
-   if (fb->bo)
+   if (surface && fb->bo)
       gbm_surface_release_buffer(surface, fb->bo);
 
    fb->bo = NULL;
@@ -81,9 +82,11 @@ page_flip_handler(int fd, unsigned int frame, unsigned int sec, unsigned int use
    struct wlc_backend_surface *bsurface = data;
    struct drm_surface *dsurface = bsurface->internal;
 
-   uint8_t next = (dsurface->index + 1) % NUM_FBS;
-   release_fb(dsurface->surface, &dsurface->fb[next]);
-   dsurface->index = next;
+   if (!drm.use_egldevice) {
+      uint8_t next = (dsurface->index + 1) % NUM_FBS;
+      release_fb(dsurface->gbm_surface, &dsurface->fb[next]);
+      dsurface->index = next;
+   }
 
    struct timespec ts;
    ts.tv_sec = sec;
@@ -107,7 +110,7 @@ drm_event(int fd, uint32_t mask, void *data)
 }
 
 static bool
-create_fb(struct gbm_surface *surface, struct drm_fb *fb)
+create_gbm_fb(struct gbm_surface *surface, struct drm_fb *fb)
 {
    assert(surface && fb);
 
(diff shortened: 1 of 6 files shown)


src/platform/backend/drm.c
```diff
@@ -141,19 +144,75 @@ create_fb(struct gbm_surface *surface, struct drm_fb *fb)
...
+   uint8_t *map;
+   uint32_t fd;
+
+   struct drm_mode_create_dumb create_request = { 0 };
+   create_request.width = width;
+   create_request.height = height;
+   create_request.bpp = 32; /* RGBX8888 */
+
+   int ret;
+   ret = drmIoctl(drm.fd, DRM_IOCTL_MODE_CREATE_DUMB, &create_request);
```
Cloudef: You can just directly check the return values here. Temporary ret serves no purpose.

src/platform/backend/drm.c
```diff
@@ -239,11 +307,13 @@ add_output(struct gbm_device *device, struct gbm_surface *surface, struct drm_ou
...
    dsurface->crtc = info->crtc;
-   dsurface->surface = surface;
+   dsurface->gbm_surface = surface;
    dsurface->device = device;
 
+   bsurface.use_egldevice = drm.use_egldevice;
+   bsurface.drm_fd = drm.fd;
    bsurface.display = (EGLNativeDisplayType)device;
-   bsurface.display_type = EGL_PLATFORM_GBM_MESA;
+   bsurface.display_type = drm.use_egldevice ? EGL_PLATFORM_DEVICE_EXT : EGL_PLATFORM_GBM_KHR;
```
Cloudef: enclose the terninary into ().

rubdos: Why is that? I never heard about that. Just curious.

Enerccio: it has to do with priority, I mean here it does not matter, but generally `a ? 3 : 4 +1` is different from `(a ? 3 : 4) + 1`

Cloudef: @rubdos Just style, it has no technical reason unless you need to consider priority.

src/platform/backend/drm.c
```diff
@@ -487,6 +571,8 @@ update_outputs(struct chck_pool *outputs)
 bool
 wlc_drm(struct wlc_backend *backend)
 {
+   const char *use_egldevice = getenv("WLC_USE_EGLDEVICE");
+   drm.use_egldevice = !chck_cstr_is_empty(use_egldevice);
```
Cloudef: Maybe consider https://github.com/Cloudef/wlc/blob/f4ff049cc30242eec9f33af7129d2e2a2865fb5a/src/platform/render/gles2.c#L910 this way instead rather than just checking env var has something set.

src/platform/context/egl.c
```diff
@@ -119,6 +163,64 @@ has_extension(const struct ctx *context, const char *extension)
...
+      const char *device_extensions = eglQueryDeviceStringEXT(devices[i], EGL_EXTENSIONS);
+      if (is_extension_supported(device_extensions, "EGL_EXT_device_drm")) {
+         device = devices[i];
+         break;
+      }
+   }
+
+   free(devices);
+
+   return (device == EGL_NO_DEVICE_EXT) ? NULL : device;
```
Cloudef: (device == EGL_NO_DEVICE_EXT ? NULL : device)

src/platform/context/egl.c
```diff
@@ -119,6 +163,64 @@ has_extension(const struct ctx *context, const char *extension)
...
+   if (!is_extension_supported(extensions, "EGL_EXT_device_base") &&
+       (!is_extension_supported(extensions, "EGL_EXT_device_enumeration") ||
+        !is_extension_supported(extensions, "EGL_EXT_device_query")))
+      return NULL;
+
+   ret = eglQueryDevicesEXT(0, NULL, &num_devices);
+   if (!ret || num_devices < 1)
+      return NULL;
+
+   devices = calloc(sizeof(EGLDeviceEXT), num_devices);
```
Cloudef: I think rather than allocating, we can have some sensible upper limit like 255 or something. I doubt there is ever that many devices.

src/platform/context/egl.c
```diff
@@ -170,25 +272,71 @@ terminate(struct ctx *context)
...
+      }
 
-   /* In practise any EGL 1.5 implementation would support the EXT extension */
-   if (has_extension(context, "EGL_EXT_platform_base")) {
-      PFNEGLGETPLATFORMDISPLAYEXTPROC getPlatformDisplayEXT =
-            (void *) eglGetProcAddress("eglGetPlatformDisplayEXT");
-      if (getPlatformDisplayEXT)
-          return getPlatformDisplayEXT(type, native, NULL);
+      return context->api.eglGetPlatformDisplayEXT(type, native, NULL);
    }
```
Cloudef: @jwrdegoede Does these changes break anything introduced in #233?

jwrdegoede: No I don't think so, all the checks are still there just moved to a different place.

Cloudef:
Overall looks good. Maybe add documentation in README as well.

Timidger:
Tried testing this on my machine with both Way Cooler and Sway, got the same error from wlc:

```
Failed to activate vt1 for restoration
Falide to switch back to vt1
```

I made sure I installed the drivers properly, because I can use Gnome + Xorg properly with this driver loaded. Using Nvidia gtx1070 card, not that it probably matters

4e554c4c:
@Timidger Do you have Nvidia's DRM KMS module enabled? It is unstable and disabled by default, but needed for EGL. Use `modprobe -r nvidia-drm ; modprobe nvidia-drm modeset=1` if you don't.

vially:
@Cloudef I've addressed your comments and pushed the changes.

I noticed a very strange thing though. It seems I can still run the `example` using the `GBM` buffer API instead of the `EGL streams` on the NVIDIA proprietary driver but it has very poor performance. Any idea how this might be possible given that the proprietary NVIDIA driver never advertised that they support the `GBM` API?

Timidger:
@4e554c4c Thanks for the hint, I did have that disabled...but now it seems I have a screwy setup since it keeps segfaulting and after checking the core dump it seems it's trying to use nouveau still....oh well it's about time I switched distros on this machine anyways :P. Thanks for all your help.

Cloudef:
@vially I'm not sure. Maybe it fallbacks to some software implementation. Do you happen to have gbm compatible gpu in your pc also?

vially:
@Cloudef Yes, I do have an Intel CPU with integrated graphics. Although `/dev/dri` only lists one card and the monitor is connected to the NVIDIA HDMI port so I'm not sure how that would work. Strange...

@Timidger I've tried running the `example` on my main work PC which is a multi-monitor setup driven by a 970 GTX and it seems to crash as well. I'll try to look more closely into it to see what might be the problem (although it seems to work just fine on my other PC which has a single monitor and a 750 GTX card). I remember from some early test runs that all outputs where assigned the same CRTC id, but I'm not sure if that's still the case using the latest code. I'll have to try and see.

Cloudef:
Lets merge and see what happens.

atomnuker:
@vially are you happy that you've now contributed to the segmentation of the entire Linux desktop? By merging this no problem has been solved, but you've worsened an existing one. Because Nvidia couldn't admit they weren't the first and didn't want to use an existing, nice API created by Intel they NIH'd this one, and now it falls onto projects to support both? This is absolute crap, and instead of yielding to Nvidia you should have send them an angry email like this one to support GBM. Or better yet a PR, though don't count on it being merged because they'd rather NIH it.
Now instead of only 1 compositor implementing this piece of crap we have 2, and that's alot more than 0.

vially:
@atomnuker I'm sorry you feel this way. I don't like the current situation either with having two competing buffer APIs. But I'm also trying to be pragmatic and I thought it might be useful if we had `wlc` working on proprietary nVidia drivers until the nVidia and Wayland developers decide on a single buffer API.

I also happen to own an nVidia card which is not supported by the nouveau driver so I really didn't have any other option if I wanted to use Wayland.

leigh123linux:
@vially 

Are you using glvnd enabled mesa?, if so /usr/share/glvnd/egl_vendor.d/50_mesa.json overrides /usr/share/glvnd/egl_vendor.d/10_nvidia.json

Renaming 10_nvidia.json to 90_nvidia.json should correct it.

> I noticed a very strange thing though. It seems I can still run the example using the GBM buffer API instead of EGL streams on the NVIDIA proprietary driver but it has very poor performance. Any idea how this might be possible given that the proprietary NVIDIA driver never advertised that they support the GBM API?

vially:
@leigh123linux Thanks, I wasn't aware of that. I'll experiment with that to see if it makes any difference.

Issue #247, opened 2017-04-04
RFC: Buffer API autodetection


vially:
Currently `wlc` always chooses `GBM` as the default buffer API unless the `WLC_USE_EGLDEVICE` environment variable is set to a truthy value in which case the `EGL streams` buffer API is chosen.

This means that when running `wlc` under the proprietary nVidia drivers **you have to** set the `WLC_USE_EGLDEVICE` environment variable before running `wlc`. While not a big issue, this is a little inconvenient and it's always nice when a program runs without having to do any configuration whatsoever.

Would you be open to accept a PR which:
- removes the `WLC_USE_EGLDEVICE` environment variable config option
- adds `WLC_BUFFER_API` environment variable config option with possible values `GBM` and `EGL` which forces the buffer API to the given type
- implements some form of auto-detection for the buffer API (I'm thinking something simple like checking if the  `nvidia` kernel module is loaded, similar to the way [sway does it](https://github.com/SirCmpwn/sway/blob/7d43a76b4e765eb8072c09cdec3847e877cf65d7/sway/main.c#L65), but other suggestions are welcome) so that wlc would run on proprietary nVidia drivers without any configuration

Cloudef:
Sounds ok. Not sure about the auto-detection. It's not hard to export WLC_BUFFER_API=EGL in env. But if you think it is worth it, go for it.

SirCmpwn:
:+1: for auto-detection or we're gonna get a truckload of annoying support requests

vially:
@Cloudef  Great. In that case I'll try to prepare a PR tomorrow if I'll have time.

@SirCmpwn If you're okay with it I'll mostly just copy-paste and adapt your [detect_proprietary](https://github.com/SirCmpwn/sway/blob/7d43a76b4e765eb8072c09cdec3847e877cf65d7/sway/main.c#L55) function. But isn't there a `C` function somewhere in the standard library which does what [read_line](https://github.com/SirCmpwn/sway/blob/7d43a76b4e765eb8072c09cdec3847e877cf65d7/common/readline.c#L6) does so I don't have to copy that too? (I'm asking because I don't know `C` and that function seems generic enough that **it might be** in the standard library).

SirCmpwn:
No, there's not.

Drakulix:
If you are going to do auto-detection you should do it the right way.

[...]

PR #248, opened 2017-04-05
Implement buffer API autodetection

vially:
Implement buffer API auto-detection as discussed in #247 

For the auto-detection logic I've used the `drmGetVersion` function which seems to return the driver used by the DRM device.

### DIFF
--- a/src/platform/backend/drm.c
+++ b/src/platform/backend/drm.c
@@ -591,10 +591,26 @@ update_outputs(struct chck_pool *outputs)
    return count;
 }
 
+static bool
+use_egldevice(int drm_fd)
+{
+   const char *buffer_api = getenv("WLC_BUFFER_API");
+   if (chck_cstr_is_empty(buffer_api)) {
+      drmVersion *version;
+      if (!(version = drmGetVersion(drm_fd)))
+         return false;
+
+      bool use_egl = chck_cstreq(version->name, "nvidia-drm");
+      drmFreeVersion(version);
+      return use_egl;
+   }
+
+   return chck_cstreq(buffer_api, "EGL");
+}
+
 bool
 wlc_drm(struct wlc_backend *backend)
 {
-   chck_cstr_to_bool(getenv("WLC_USE_EGLDEVICE"), &drm.use_egldevice);
    drm.fd = -1;
 
    const char *device = getenv("WLC_DRM_DEVICE");
@@ -610,6 +626,8 @@ wlc_drm(struct wlc_backend *backend)
 
    if (drm.fd < 0)
       goto card_open_fail;
+
+   drm.use_egldevice = use_egldevice(drm.fd);
 
    if (!drm.use_egldevice) {
       /* GBM will load a dri driver, but even though they need symbols from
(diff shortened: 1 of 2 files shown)


src/platform/backend/drm.c
```diff
@@ -1,6 +1,7 @@
 #include <assert.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
```
Cloudef: We use _DEFAULT_SOURCE so, string.h should be enough.

src/platform/backend/drm.c
```diff
@@ -591,10 +592,24 @@ update_outputs(struct chck_pool *outputs)
...
+   const char* bufferApi = getenv("WLC_BUFFER_API");
+   if (chck_cstr_is_empty(bufferApi)) {
+      drmVersion *version = drmGetVersion(drmFd);
+      bool use_egl = chck_cstreq(version->name, "nvidia-drm");
+      drmFreeVersion(version);
+
+      return use_egl;
+   }
+
+   return !strcasecmp(bufferApi, "egl");
```
Cloudef: Then again, I don't mind if this was just case sensitive comparsion since strcasecmp isn't standard function.

src/platform/backend/drm.c
```diff
@@ -591,10 +592,24 @@ update_outputs(struct chck_pool *outputs)
...
 
+static bool
+use_egldevice(int drmFd)
+{
+   const char* bufferApi = getenv("WLC_BUFFER_API");
+   if (chck_cstr_is_empty(bufferApi)) {
+      drmVersion *version = drmGetVersion(drmFd);
+      bool use_egl = chck_cstreq(version->name, "nvidia-drm");
+      drmFreeVersion(version);
+
```
Cloudef: Remove this empty line.

vially:
@Cloudef  Pushed another version which addressed your comments

src/platform/backend/drm.c
```diff
@@ -591,10 +591,23 @@ update_outputs(struct chck_pool *outputs)
    return count;
 }
 
+static bool
+use_egldevice(int drmFd)
```
Cloudef: Change the camelCase variable names to snake_case

src/platform/backend/drm.c
```diff
@@ -591,10 +591,23 @@ update_outputs(struct chck_pool *outputs)
...
+{
+   const char* bufferApi = getenv("WLC_BUFFER_API");
```
Cloudef: Make the star hug the variable name.

src/platform/backend/drm.c
```diff
@@ -591,10 +591,23 @@ update_outputs(struct chck_pool *outputs)
...
+   if (chck_cstr_is_empty(bufferApi)) {
+      drmVersion *version = drmGetVersion(drmFd);
```
Cloudef: Can the function return NULL? If not, add assert(version); below.

vially:
Pushed another version which addressed latest feedback.

Cloudef:
Looks ok to me, thanks.

PR #253, opened 2017-04-14
Exposed api to grab pointer texture

Enerccio:
Provides a way to get either custom texture surface from client or default texture from render

### DIFF
--- a/include/wlc/wlc-render.h
+++ b/include/wlc/wlc-render.h
@@ -20,7 +20,7 @@ extern "C" {
 
 /** Allowed pixel formats. */
 enum wlc_pixel_format {
-   WLC_RGBA8888,
+   WLC_RGBA8888
 };
 
 /**
@@ -69,6 +69,7 @@ enum wlc_surface_format {
     SURFACE_Y_UV,
     SURFACE_Y_U_V,
     SURFACE_Y_XUXV,
+	 INVALID,
 };
 
 /**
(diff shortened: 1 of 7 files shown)


Cloudef:
What exactly is the use case for this?

Enerccio:
well if you render everything yourself, you do need a way to grab the surface of the pointer (and the default render texture)

Enerccio:
Personally I use it here: https://github.com/Enerccio/nyarlathotep-wm/blob/master/src/render.c#L401

Cloudef:
If you want to render everything yourself. I would probably instead work on making this plugable https://github.com/Cloudef/wlc/blob/master/src/platform/render/render.c#L142-L150 so you can replace the renderer itself instead of clearing and drawing everything again.

ammen99:
@Cloudef The compositor doesn't have to render everything again, if you want to do custom rendering, you can set the output mask to 0.

Enerccio:
@Cloudef I looked at that, but that does actual rendering, but not the order of view render nor anything else. which is why I opted to actually render everything afterwards

I also tried to have custom renderer api (it's master of my fork actually) but for some reason, if I had empty render view, I would get black texture later when rendering myself

Enerccio:
Here is the exposed api for render but I am not sure if it is in state to PR: https://github.com/Enerccio/wlc/

PR #255, opened 2017-04-15
Small changes to make working with Xwayland easyier

kozec:
In my long running attempt to create more traditional WM based on WLC, I found two things to really hard to do outside of wlc:

- Native wayland applications are drawing decoration for themselves, but X11 window has to be decorated by WM/Compositor. This PR adds `WLC_BIT_X11` and `WLC_BIT_BORDERLESS` view types to help deciding which is which without doing complicated stuff to determine something that WLC already knows.
- Closing X11 windows means connecting to Xserver, querying, sending requests, calling XKillWindow... While WLC already does this internally. This PR adds API to do all that using already existing wlc->x11 connection.

### DIFF
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -56,6 +56,7 @@ enum wlc_view_type_bit {
    WLC_BIT_MODAL = 1<<3, // Modal windows (x11)
    WLC_BIT_POPUP = 1<<4, // xdg-shell, wl-shell popups
    WLC_BIT_X11 = 1<<5, // Any x11 window
+   WLC_BIT_BORDERLESS = 1<<6, // Borderless (undecorated) window (x11)
 };
 
 /** wlc_set_view_properties_updated_cb(); */
--- a/include/wlc/wlc.h
+++ b/include/wlc/wlc.h
@@ -449,6 +449,11 @@ uint32_t wlc_view_get_type(wlc_handle view);
  */
 void wlc_x11_window_delete(uint32_t window);
 
+/**
+ * Kills xwayland window unconditionally.
+ */
+void wlc_x11_window_kill(uint32_t window);
+
 /** Set type bit. Toggle indicates whether it is set or not. */
 void wlc_view_set_type(wlc_handle view, enum wlc_view_type_bit type, bool toggle);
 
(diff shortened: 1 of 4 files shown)


Cloudef:
Why are the wlc_x11_ functions needed for public?

kozec:
They are not exactly needed, but it's something that's nice to have. It allows to implement "Kill window" option in window menu without creating separate connection to X server. But if it means problem, I can create PR without them.

Enerccio:
why not have them public though? There is nothing wrong with having as detailed public api as possible, for having options ready for people using wlc? Surely that is not a bad thing... if cluttering is a point, just move them to separate api header?

Timidger:
But if wlc always does this internally when you close a `WlcView`, then why bother exposing it? What use case is there for the compositor to go around wlc and send the kill command to the X server itself when wlc already handles it?

If it's not necessary, I don't see why it should be made public.

kozec:
> But if wlc always does this internally when you close a WlcView, then why bother exposing it? What use case is there for the compositor to go around wlc and send the kill command to the X server itself when wlc already handles it?

My main use case is this kind of dialog:
cantclose

It is supposed to appear when X11 window claims to support WM_DELETE_WINDOW, but doesn't react to event (usually because its process hanged). Currently there is no way to get rid of such window using just WLC api.

Other use case may be allowing user to kill window instead of closing it, as some older WMs does.

kozec:
@Cloudef sorry to bother you, but I really would like your option on this. Does it looks at least theoretically mergeable? It looks like I'll have to do more crazy stuff with X11 windows (for example, I'm looking at enabling borders on menus right now) and I'm not sure in which direction should I go.

Would it be better to add even more methods to wlc, or scrap even those in this commit and expose XIDs and display connection, so library user can do whatever he wants without help of wlc?

Cloudef:
Preferably, you would not need to do anything x11 related. But if it's really needed just expose the display and xids in wlc-xwayland.h or something.

Cloudef:
The WLC_BIT_BORDERLESS is okay.

kozec:
> But if it's really needed just expose the display and xids in wlc-xwayland.h or something.

Will do, thanks :)

> The WLC_BIT_BORDERLESS is okay.

I'll create another PR just with that for time being.

PR #256, opened 2017-04-16
added pre/post swap render events

Enerccio:
cause everyone loves more events

technically because I do custom rendering and post render hook was before pointer render, which caused blinking cursor, so I added two more events (pre/post swap) so I can use pre swap for my custom rendering


Cloudef:
Is post_swap useful for anything?

Enerccio:
added it for sake of completeness but not to me right now at least

Enerccio:
thinking about it, it might be useful when rendering was redirected and we want to know when swap happened?

PR #257, opened 2017-04-17
Changes to pointer logic, more api for working with surfaces and render repaint callback + changes related

Enerccio:
* added api to get pointer tip
* added api to get pointer surface
* added api to bypass output repaint loop in case you want to repaint yourself (like I do)


Enerccio:
merged into https://github.com/Enerccio/wlc/tree/experimental instead

PR #259, opened 2017-04-18
Events for minimization

Enerccio:
Surprised it wasn't there already, everything was set up backend, just callback and getter was missing


Cloudef:
Squash the commits.

Enerccio:
squashed

PR #265, opened 2017-05-08
Add modesetting support

zsrkmyn:
I've made some effort to make wlc to support modesetting due to the issue I posted SirCmpwn/sway#1212, but the codes is not fully tested, since I am not familiar with wlc and wayland. Frankly speaking, It's just my first try to use wayland.

### DIFF
--- a/example/example.c
+++ b/example/example.c
@@ -299,6 +299,30 @@ pointer_motion(wlc_handle handle, uint32_t time, const struct wlc_point *positio
    return (compositor.action.view ? true : false);
 }
 
+static bool
+output_created(wlc_handle output)
+{
+   const char *name = wlc_output_get_name(output);
+
+   size_t nmodes;
+   const struct wlc_output_mode *modes = wlc_output_get_modes(output, &nmodes);
+   printf("Output %s created, %zu modes found:\n", name, nmodes);
+
+   for (size_t i = 0; i < nmodes; i++) {
+      printf("mode %zu: %ux%u@%u\n", i, modes[i].width, modes[i].height, modes[i].refresh);
+   }
+   if (!strcmp(name, "LVDS-1")) {
+      size_t idx = (size_t)-1;
+      for (idx = 0; idx < nmodes; idx++) {
+         if (modes[idx].width == 1366)
+            break;
+      }
+      printf("Setting mode to %ux%u\n", modes[idx].width, modes[idx].height);
+      wlc_output_set_mode(output, idx);
+   }
+   return true;
+}
+
 static void
 cb_log(enum wlc_log_type type, const char *str)
 {
@@ -312,6 +336,7 @@ main(void)
    wlc_log_set_handler(cb_log);
 
    wlc_set_output_resolution_cb(output_resolution);
+   wlc_set_output_created_cb(output_created);
    wlc_set_view_created_cb(view_created);
    wlc_set_view_destroyed_cb(view_destroyed);
    wlc_set_view_focus_cb(view_focus);
--- a/example/example.c
+++ b/example/example.c
@@ -299,30 +299,6 @@ pointer_motion(wlc_handle handle, uint32_t time, const struct wlc_point *positio
    return (compositor.action.view ? true : false);
 }
 
-static bool
-output_created(wlc_handle output)
-{
-   const char *name = wlc_output_get_name(output);
-
-   size_t nmodes;
-   const struct wlc_output_mode *modes = wlc_output_get_modes(output, &nmodes);
-   printf("Output %s created, %zu modes found:\n", name, nmodes);
-
-   for (size_t i = 0; i < nmodes; i++) {
-      printf("mode %zu: %ux%u@%u\n", i, modes[i].width, modes[i].height, modes[i].refresh);
-   }
-   if (!strcmp(name, "LVDS-1")) {
-      size_t idx = (size_t)-1;
-      for (idx = 0; idx < nmodes; idx++) {
-         if (modes[idx].width == 1366)
-            break;
-      }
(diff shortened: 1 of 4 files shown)


example/example.c
```diff
@@ -299,6 +299,30 @@ pointer_motion(wlc_handle handle, uint32_t time, const struct wlc_point *positio
...
+   const char *name = wlc_output_get_name(output);
+
+   size_t nmodes;
+   const struct wlc_output_mode *modes = wlc_output_get_modes(output, &nmodes);
+   printf("Output %s created, %zu modes found:\n", name, nmodes);
+
+   for (size_t i = 0; i < nmodes; i++) {
+      printf("mode %zu: %ux%u@%u\n", i, modes[i].width, modes[i].height, modes[i].refresh);
+   }
+   if (!strcmp(name, "LVDS-1")) {
```
SirCmpwn: This code seems too specific to your system. Can you remove it?

include/wlc/wlc.h
```diff
@@ -159,6 +159,13 @@ struct wlc_modifiers {
    uint32_t leds, mods;
 };
 
+/** Dispaly mode settings **/
+struct wlc_output_mode {
+   int32_t refresh;
+   int32_t width, height;
+   uint32_t flags; // WL_OUTPUT_MODE_CURRENT | WL_OUTPUT_MODE_PREFERRED
```
SirCmpwn: Can you define a WLC enum for this? The values can match the wayland enum's values.

include/wlc/wlc.h
```diff
@@ -364,6 +371,15 @@ bool wlc_output_set_views(wlc_handle output, const wlc_handle *views, size_t mem
 /** Focus output. Pass zero for no focus. */
 void wlc_output_focus(wlc_handle output);
 
+/** Get output modes */
+const struct wlc_output_mode* wlc_output_get_modes(wlc_handle output, size_t *out_memb);
```
SirCmpwn: Presumably this structure has a limited lifetime of validity. Can you speak to that in the docs?

SirCmpwn: "The docs" here just referring to the comment here in the header file

src/compositor/output.c
```diff
@@ -600,6 +600,25 @@ wlc_output_set_backend_surface(struct wlc_output *output, struct wlc_backend_sur
    return false;
 }
 
+static void
+wlc_output_update_mode(struct wlc_output *output)
+{
+
+   bool set_resolution = false;
+
+   {
```
SirCmpwn: What's with the extra scope here?

SirCmpwn: Oh, I guess you just moved this code from elsewhere. Still, would remove the braces in this case.

SirCmpwn:
If I understand correctly, the mode change should be accompanied with a call to wlc_output_set_resolution, right?

zsrkmyn:
Thanks for patiently reviewing the code. I'll fix the problems you metioned above soon.

I am not sure whether mode change should be accompainied with a call to wlc_output_set_resolution or not. The original code does this so I keep it unchanged. Should I split it out?

SirCmpwn:
wlc_output_set_resolution is used to change the size of wlc's internal buffer (which is distinct from the mode).

SirCmpwn:
I'm fine either way with calling it manually or having wlc do it on set_mode. Should be documented, though.

zsrkmyn:
It seems the [code here](https://github.com/Cloudef/wlc/blob/master/src/platform/backend/drm.c#L213) related to drm fails when changing resolution. It says `No space left on device`. I may do some research on drm tomorrow.

zsrkmyn:
It's kind of hard to solve the problem. The problem occurs because the width and the height of `gbm_surface` (more spcifically, the width and the height of the frame buffer in `gbm_surface`) mismatch with the resolution. However, the `gbm_surface` (and its frame buffer) is created when the output created (drm.c:L584), with the width and height set to `info->width` and `info->height`, and these two variables are set in `drm_query()` to the current mode with highest priority (drm.c:L467), and to the preferred mode with second highest priority  (drm.c:L476). So, in my case, they were set to current mode, which is 1280x720, so the size of the frame buffer in `gdm_surface` is ceated with the size of 1280x720. And when we change the mode to 1920x1080, the frame buffer is unchanged, and thus when the `page_flip()` is called, the error occurs.

Since the `gbm_surface` is created when the output is created and is backend-related (only to drm), it's not easy to fix the problem. I am now trying to add a new api to bsurface.api so that surface can be update via the api after setting mode, but the original code of managing frame buffer seems a bit complicated for me.

Enerccio:
wouldn't easiest solution be to have output carry a "backend resize" internal callback that then would be called when modeset changes? Then gbm could have code that would destroy old frame buffer and create new one?

SirCmpwn:
FYI, sway 0.14-rc1 shipped today, so the sway side of modesetting support is not happening. I still want to see this ship, but I ~~forgot about this PR~~ prefer a conservative release for now. Will try to make sure it's merged around 0.15 time (or before).

zsrkmyn:
I'm so sorry to put off fixing bugs in this PR for long, but I am really busy these days and almost forget this PR as well... I wonder whether sway still depends on wlc's modesetting. If so, I may try to work on it next week.

SirCmpwn:
No worries. It can wait until the next release cycle. I'll have a look at it myself then if need be.

SirCmpwn:
Was looking into this for sway 0.15 and I decided that the effort required to clean this up and ship it is greater than the amount I'm willing to put in considering that wlroots is on the horizon. Sorry - if you want to clean this up yourself though I'd be happy to merge it!

zsrkmyn:
I am glad that sway has supported modesetting, and I am so sorry that I finally decide to give up this PR. Thank you a lot for all your assistance!

Issue #266, opened 2017-05-22
Touch event reports view by mouse position


kozec:
Hello,

I'm playing with touch events and I've noticed pretty weird behavior - view reported by touch events is not view that user is touching, but view that mouse pointer is above.

Is this intentional?

SirCmpwn:
The mouse pointer should probably be moved in response to touch events.

kozec:
> The mouse pointer should probably be moved in response to touch events.

Yes, but even doing that is not enough, as first touch event always fires before pointer is moved and this reports wrong view.

SirCmpwn:
Sure, it sounds like a bug to me.

kozec:
By the way, another problem where I'm not 100% sure if it's a bug...

[Code in touch.c](https://github.com/Cloudef/wlc/blob/c263cc3d064ba78da9c8728722432717ed9385c0/src/compositor/seat/touch.c#L78) appears to send touch events with coordinates in screen space, but I'm pretty sure that Wayland client expects them in surface space. Tapping on gnome calculator suggest the same, as button that is actually pressed when I touch its window is offseted and offset depends on window position.

SirCmpwn:
Definitely sounds like a bug to me.

PR #267, opened 2017-05-22
Change way how view paramater in touch event is computed

kozec:
As I've described in #266, wlc currently reports touch events with view set to one over which mouse cursor is hovering.

With this PR, touch would remember view that was touched initially, so touch down event and subsequent touch events are reported for that view.


SirCmpwn:
Thanks!

PR #272, opened 2017-07-01
Notify focused client on changed selection

nyorain:
This fixes a data.c bug. Currently, it does not send notifications to the focused client when a selection changes which results in a segmentation fault if the clients tries then to receive from the (no longer valid) data offer. Usually clients don't use the wayland protocol when they want to access their own data source but manage it internally which is why this did not introduce any real problems until now, but would e.g. when implementing Sircmpwn/sway#926 (since then the source is changed without that the client can know it).

### DIFF
--- a/src/compositor/seat/data.c
+++ b/src/compositor/seat/data.c
@@ -6,6 +6,7 @@
 #include <chck/string/string.h>
 #include "visibility.h"
 #include "macros.h"
+#include "compositor/view.h"
 #include "internal.h"
 #include "macros.h"
 #include "data.h"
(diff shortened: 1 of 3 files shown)


SirCmpwn:
This looks fine, but I'm not convinced it'll solve https://github.com/SirCmpwn/sway/issues/926. CLI clients (i.e. clients that don't open a window) aren't going to get focus.

nyorain:
I fixed the sway issue with a new ipc message (currently preparing the pr).

SirCmpwn:
Cool, thanks!

PR #273, opened 2017-07-06
Public api to receive clipboard data

nyorain:
This enables compositors using wlc to receive the supported mime types of the current selection and then request the data in a certain format. It also shows the usage in the example compositor, currently using the ctrl+comma shurtcut to output the current text data selection (there might be something better for this, didn't want to use ctrl+v since this is usually used by applications, not sure if it matters in this case).
It also fixes various smaller issues, see the second commit messag for details.
The public wlc api was designed to offer the greatest possible freedom, which makes it a bit harder to use  (see the example, one has to call pipe2 in the compositor and add a wlc event source).
Could be used to (fully) fix SirCmpwn/sway#926.


SirCmpwn:
pipe2 is a GNU extension.

nyorain:
right, should be working with ```pipe``` as well.

SirCmpwn:
Thanks!

PR #279, opened 2017-09-13
Add elogind support

Dudemanguy911:
During configuration, wlc only enables logind support if systemd itself is present. This adds a check for elogind so that way systems without systemd can still take advantage of user sessions with logind if they have elogind installed.


SirCmpwn:
Can you add the email you committed this with to your GitHub account?

Dudemanguy911:
Sure, added it just now.

SirCmpwn:
Thanks!

PR #280, opened 2017-10-02
Fix xwm selection release

alexander-smoktal:
In case xwayland terminates before compositor activated, xwm selection listener link is invalid. This causes crash in `wlc_xwm_selection_release`.

Fixed as in `wlc_pointer_release` with notification callback check.


SirCmpwn:
Thanks!

PR #282, opened 2017-10-14
Add drag and drop stub

Laaas:
This among other things allows SDL to work on wlc, since it requires version 3. Wlc only provided version 2 before this.

Some useless stuff is done in the set_actions functions currently, but this useless stuff could potentially be expanded into a functioning dnd system.


SirCmpwn:
Sweet, thanks!

PR #286, opened 2017-12-28
Don't send mode info twice on DRM refresh

Timidger:
Similar fix as [this issue in wlroots](https://github.com/swaywm/wlroots/issues/535)

I don't know this codebase very well, so this might not be the most correct place to fix it. I saw a similar structure to how it works in wlroots and it seems to work on my local copy.


SirCmpwn:
Thanks!

